PR #48, opened 2017-06-08
Add --userlist (-U) param for debugging

Thing342:
Adds a --userlist param that sets which users' list files to read in (defaulting to all users when not set)

Example:
`./siteupdate.py ... --userlist oscar mapcat froggie ...` would only process oscar, mapcat, and froggie's LIST files.

(includes some formatting as well)


jteresco:
Looks good, thanks for the contribution.

PR #102, opened 2018-09-08
new program to filter tm-master.nmp by region

yakra:
We don't necessarily have to call it "nmpfilter", if anyone has anything better to call it. ("nmpregions" or something for example, maybe?)
OTOH, what else could there be that we'd want to filter tm-master.nmp by?

### DIFF
--- a/nmpfilter/README.md
+++ b/nmpfilter/README.md
@@ -1,0 +1,10 @@
+# nmpfilter
+
+**Purpose:**<br>
+Splits *tm-master.nmp* into smaller .nmp files filtered by region.
+
+**usage:**<br>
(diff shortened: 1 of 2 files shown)


yakra:
>  OTOH, what else could there be that we'd want to filter tm-master.nmp by?
* filter by country
* filter out FPs
* filter in LI entries

Stay tuned!

jteresco:
Filtering types of NMPs might be better to do in HDX by a couple checkboxes to show/hide things that are green/yellow/red, but I have no objection to creating such filters with the nmpfilter or other programs that will create smaller files with only what you're looking to see.

Issue #57, opened 2017-07-30
Error check for wpt line format


jteresco:
These lines got into mn.mn298.wpt, which caused the site update to fail:

```
+x01 http://www.openstreetmap.org/?lat=19/44.28690&lon=-93.25970
+x02 http://www.openstreetmap.org/?lat=19/44.28646&lon=-93.25973
+x03 http://www.openstreetmap.org/?lat=19/44.28644&lon=-93.25929
MottAve http://www.openstreetmap.org/?lat=19/44.28694&lon=-93.25901
```
Unfortunately, the site update did not crash until during DB update, leaving the site offline.

Need to add in checks for when a wpt is read and the lines cannot be parsed properly and flag those as errors earlier in the update process.

ajfroggie:
Thought I caught those....oops.  Downside of our format and how it differs from default OSM URLs...

ajfroggie:
Are you fixing those or should I?

jteresco:
Just pushed up the fixed file.  Fortunately just lines in that one file had the error.

yakra:
A .wpt line containing only whitespace crashes siteupdate.py. (A newline only, sans whitespace, is OK.)
siteupdate.py stops reading in remaining Waypoints from the offending, or future, .wpt files, proceeds thru writing graphs (of this partial dataset), and crashes when trying to perform the first of the main datachecks.

yakra:
> Unfortunately, the site update did not crash until during DB update, leaving the site offline.

Interesting. I made a small program to test this, and it crashed immediately upon attempting the string->float conversion with bad input. If you indeed got different results, I'll just chalk it up to more differences between the FreeBSD & Ubuntu flavours of Python.

---
**Current code:**
https://github.com/TravelMapping/DataProcessing/blob/dc8a05c0968abaee8b136305c5e5118fdcdbb84c/siteupdate/python-teresco/siteupdate.py#L267-L272
Execution time:
* 104.775 s single-threaded, 115.95 s multi-threaded on noreaster
* 127.85 s single-threaded, 191.025 s multi-threaded on BiggaTomato
---
**Option 1:**
```Python
        # last has the URL, which needs more work to get lat/lng
        url_parts = parts[-1].split('=')
        lat_string = url_parts[1].split("&")[0] # chop off "&lon"
        lng_string = url_parts[2].split("&")[0] # chop off possible "&zoom"

        # make sure lat_string is valid
        point_count = 0
[...]

PR #125, opened 2018-11-06
remove LABEL_NO_VALID

yakra:
`LABEL_NO_VALID` only appears in this comment listing the error code string values, but nowhere in the actual code.
It appears to be a placeholder for the old CHM `label...has no valid characters`, as described by Tim:
> **7. label...has no valid characters:** Always a true error. Waypoint labels must have a nonzero number of valid characters.

I see no need to give this its own separate datacheck; these cases will be handled by LABEL_INVALID_CHAR. If there's a label at all, it must have 1 or more characters. If there are no valid characters, then there are 1 or more invalid characters.

Can a label have no characters at all?
* If a .wpt line contains only one field & no whitespace, this text is parsed as both the label **and** the URL *(a form of graceful degradation IMO)*.
* If a .wpt line is blank (only an extra newline, no whitespace), it's ignored, and never hits the Waypoint constructor. Good!
* If a .wpt line contains only whitespace, siteupdate.py stops reading in remaining Waypoints from the offending, or future, .wpt files, proceeds thru writing graphs (of this partial dataset), and crashes when trying to perform the first of the main datachecks. I haven't yet looked into *how* this happens, but...

This is more an issue of robustness with Waypoint init (#57), and doesn't make it to the point of being a datacheck.

Let's say that `LABEL_NO_VALID` is out.


jteresco:
I merged all of these in after starting tonight's update.  They'll be in use for tomorrow if I remember to pull them down to noreaster before I launch one tomorrow.

jteresco:
Looks OK from my view.

PR #128, opened 2018-11-13
remove OUT_OF_BOUNDS from datacheck_always_error

yakra:
< -180° or > 180° should be allowed in some limited circumstances.
Any road spanning +/-180° (E.G., West Coast Road, Taveuni, Fiji) will get messed up if these coords are "repaired": Leaflet will draw long E-W lines spanning almost the entire globe.

### DIFF
--- a/siteupdate/python-teresco/siteupdate.py
+++ b/siteupdate/python-teresco/siteupdate.py
@@ -3007,13 +3007,9 @@ for h in highway_systems:
 
         # look for hidden termini
         if r.point_list[0].is_hidden:
-            labels = []
-            labels.append(r.point_list[0].label)
-            datacheckerrors.append(DatacheckEntry(r,labels,'HIDDEN_TERMINUS'))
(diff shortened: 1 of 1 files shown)


jteresco:
Such points would fall outside the quadtree, so maybe its initial E/W bounds should be a tad beyond 180.

yakra:
*Where to put them, though?*
I don't see the bounds as causing any harm. This would behave just as before #126, when longitude bounds were set at +/-90°, just on a much smaller scale: Quadtree Nodes along the E & W edges of the map will still receive points outside their nominal bounds, and pass them down to their child nodes based on where they fall relative to `mid_lng`.
Nodes may fill up & refine a tad faster, as they have more width in which to receive points.
But there should be relatively few (if any) of these points, and the effects should be minimal. Nothing like the long narrow "stripes" that had to contend with taking in all the waypoints for an extra 90° longitude.

PR #129, opened 2018-11-13
order pointsinuse.log, unusedaltlabels.log asciibetically

yakra:


michihdeu:
👍 Could you sort the other log files too? unusedaltlabels et cetera.

yakra:
@michihdeu you read my mind!
The others are a work in progress.

michihdeu:
@yakra http://travelmapping.net/logs/unusedaltlabels.log is not asciibetically

yakra:
This was only meant to refer to the order of labels listed within the individual lines for each route, and not the order of the routes themselves.

yakra:
Lines in [pointsinuse.log](http://travelmapping.net/logs/pointsinuse.log) and [unusedaltlabels.log](http://travelmapping.net/logs/unusedaltlabels.log) are sorted by the same criteria: by highway system, then by CSV route order within each system. I'd prefer to leave that as it already is.

Personally, as far as usability goes, when I use pointsinuse.log in maintaining Highway Data, I just search for the string matching the root I want, E.G. `me.us001(`. When looking thru the list for a specific label (rather than just copying & pasting into WPTedit), being asciibetized here helps, yes...

But my primary goal here has just been deterministic siteupdate log file ordering, so that log files will come out the same from successive runs of siteupdate.py using the same HighwayData & UserData. This way, I can get sane, human-readable DIFFs as I try out changes to siteupdate.py.

michihdeu:
Ok. I don't know what I've seen or what I thought I've seen.... Sorry. It's fine to me. I searc for pointsinuse the same way. Just use my browser search feature (FF).

PR #139, opened 2018-11-28
Sort waypoints in Quadtree

yakra:
for deterministic nearmisspoints.log & tm-master.nmp

Takes 0.8 seconds on my machine. Will be a little faster on Noreaster.
I think DIFFs are worthwhile. :)


jteresco:
I like it.  I forgot to pull this in before tonight's update, but I'll be doing so now so it's good to go for tomorrow.

Issue #88, opened 2018-05-30
Graph generation: list of systems in the graphs directory instead of hard-coded


jteresco:
https://github.com/TravelMapping/DataProcessing/blob/ec1289ae5733676d54e035dd346f04355a39e6a8/siteupdate/python-teresco/siteupdate.py#L3105

This hard-coded list should be replaced with a list of system graphs desired over in the HighwayData/graphs directory.

yakra:
Does this need to be anything more complicated than just a plain, say, **systemgraphs.txt**, reading
```
usai
usaus
cantch
eure
```

jteresco:
That's it.  I wonder if it makes sense to name it as a .csv for consistency with the others in the graphs directory, even though each line would have a single value with no separators.

jteresco:
Yes, let's include a header row, thanks.

PR #153, opened 2018-12-21
wpt line checks for #57

yakra:
@jteresco: one thing I'm not 100% certain of is the change to replace rstrip('\n') with .strip(). I want to be sure I grok how things are working behind the scenes.
* When we do a `lines = file.readlines()`, the `\n` delimiter is removed, and not contained in the lines in `lines`, correct?
* If `\n` is contained in the individual lines, does strip() get rid of it? Docs I found with some quick Googling didn't mention `\n` specifically.
* In any case, I achieved the desired effect: in addition to newline-only lines, whitespace-only lines are not passed to the Waypoint constructor or parsed at all. DIFFs of [deterministic logs](https://github.com/yakra/DataProcessing/issues/17) affected only the creation time. Siteupdate.log reported the same number of total waypoints & unique locations. I observed no unintended side effects.

### DIFF
--- a/siteupdate/python-teresco/siteupdate.py
+++ b/siteupdate/python-teresco/siteupdate.py
@@ -253,7 +253,7 @@ class Waypoint:
     root is the unique identifier for the route in which this waypoint
     is defined
     """
-    def __init__(self,line,route):
+    def __init__(self,line,route,datacheckerrors):
         """initialize object from a .wpt file line"""
         self.route = route
         parts = line.split()
@@ -266,8 +266,61 @@ class Waypoint:
             self.alt_labels = []
         # last has the URL, which needs more work to get lat/lng
         url_parts = parts[-1].split('=')
+        if len(url_parts) < 3:
+            datacheckerrors.append(DatacheckEntry(route,[self.label],'MALFORMED_URL', parts[-1]))
+            self.lat = 0
+            self.lng = 0
+            self.colocated = None
+            self.near_miss_points = None
+            return
         lat_string = url_parts[1].split("&")[0] # chop off "&lon"
         lng_string = url_parts[2].split("&")[0] # chop off possible "&zoom"
+
+        # make sure lat_string is valid
+        point_count = 0
+        for c in range(len(lat_string)):
+            # check for multiple decimal points
+            if lat_string[c] == '.':
+                point_count += 1
+                if point_count > 1:
+                    datacheckerrors.append(DatacheckEntry(route,[self.label],'MALFORMED_URL', parts[-1]))
+                    lat_string = "0"
+                    lng_string = "0"
+                    break
+            # check for minus sign not at beginning
+            if lat_string[c] == '-' and c > 0:
+                datacheckerrors.append(DatacheckEntry(route,[self.label],'MALFORMED_URL', parts[-1]))
+                lat_string = "0"
+                lng_string = "0"
+                break
+            # check for invalid characters
+            if lat_string[c] not in "-.0123456789":
+                datacheckerrors.append(DatacheckEntry(route,[self.label],'MALFORMED_URL', parts[-1]))
+                lat_string = "0"
+                lng_string = "0"
+                break
+
+        # make sure lng_string is valid
+        point_count = 0
+        for c in range(len(lng_string)):
+            # check for multiple decimal points
+            if lng_string[c] == '.':
+                point_count += 1
+                if point_count > 1:
+                    datacheckerrors.append(DatacheckEntry(route,[self.label],'MALFORMED_URL', parts[-1]))
+                    lng_string = "0"
+                    break
+            # check for minus sign not at beginning
+            if lng_string[c] == '-' and c > 0:
+                datacheckerrors.append(DatacheckEntry(route,[self.label],'MALFORMED_URL', parts[-1]))
+                lng_string = "0"
+                break
+            # check for invalid characters
+            if lng_string[c] not in "-.0123456789":
+                datacheckerrors.append(DatacheckEntry(route,[self.label],'MALFORMED_URL', parts[-1]))
+                lng_string = "0"
+                break
+
         self.lat = float(lat_string)
         self.lng = float(lng_string)
         # also keep track of a list of colocated waypoints, if any
@@ -710,9 +763,10 @@ class Route:
             file.close()
             w = None
             for line in lines:
-                if len(line.rstrip('\n')) > 0:
+                line = line.strip()
+                if len(line) > 0:
                     previous_point = w
-                    w = Waypoint(line.rstrip('\n'),self)
+                    w = Waypoint(line,self,datacheckerrors)
                     self.point_list.append(w)
                     # populate unused alt labels
                     for label in w.alt_labels:
@@ -1061,7 +1115,7 @@ class DatacheckEntry:
 
     code is the error code string, one of SHARP_ANGLE, BAD_ANGLE,
     DUPLICATE_LABEL, DUPLICATE_COORDS, LABEL_SELFREF,
-    LABEL_INVALID_CHAR, LONG_SEGMENT,
+    LABEL_INVALID_CHAR, LONG_SEGMENT, MALFORMED_URL,
     LABEL_UNDERSCORES, VISIBLE_DISTANCE, LABEL_PARENS, LACKS_GENERIC,
     BUS_WITH_I, NONTERMINAL_UNDERSCORE,
     LONG_UNDERSCORE, LABEL_SLASHES, US_BANNER, VISIBLE_HIDDEN_COLOC,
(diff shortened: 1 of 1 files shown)


yakra:
Looking OK -- TravelerList init is done similarly:
```Python
        for line in lines:
            line = line.strip()
            # ignore empty or "comment" lines
            if len(line) == 0 or line.startswith("#"):
                continue
```

> Don't know if it's the right place but `\r\n` is an issue in wpt files. I sometimes forgot to remove `\r` in the past and siteupdate crashed. Would be cool if you could catcht this. Should I submit a new issue?

I don't think there's any problem with `\r`. Lots of WPTs have been using `\r\n` extensively for years, and are parsed perfectly fine. For example, I just opened al.al108.wpt in a hex editor. DOS newlines. There are countless others, all working fine. These are common enough that we needed to get them nailed down in the very early days.
When Python does `lines = file.readlines()`, it can even cope with `\r` (hex `0D`) alone. For example, see [whopperman.list](https://github.com/TravelMapping/UserData/blob/6309036c44102eb3325d49515b32c5eef3b3cb1e/list_files/whopperman.list) (I have to wonder if whopperman initially drafted this file [on classic Mac OS](https://en.wikipedia.org/wiki/Newline#Representation) back in the day). siteupdate.py handles this just fine. GitHub has trouble displaying the line numbers. The [C++ getline function](http://www.cplusplus.com/reference/string/string/getline/) choked on this, because it's only made to work with a single delimiter; I had to find a workaround for siteupdate.cpp.

If you're able to reproduce the issue you experienced, sure, submit a new issue; describe what failed & how, and I'll see if I can reproduce it too.

michihdeu:
I remember what's the issue. Jim had another computer with a different OS and the problem only occured there. The one for tmtest?

yakra:
Interesting. @michihdeu, do you remember any discussion on either GitHub or the forum that I can read up on? (Maybe [this](https://github.com/TravelMapping/HighwayData/pull/1072#issuecomment-274592924) is it?)

> I remember what's the issue. Jim had another computer with a different OS and the problem only occured there. The one for tmtest?

ISTR that before noreaster came into service, Jim ran siteupdate.sh on his local machine (a Mac?), then transferred the files to... [looks like blizzard.teresco.org](https://github.com/TravelMapping/DataProcessing/blob/master/siteupdate/python-teresco/siteupdate.sh)?
The [wikipedia article](https://en.wikipedia.org/wiki/Newline#Representation) noted upthread lists modern macOS as using just `0A` aka `\n` for newlines.
So that's potentially consistent with what you describe -- stray `0D` aka `\r` characters, not parsed as part of a newline, left behind & fouling things up.

~I'm still mystified though.
If that HAD been a problem...~ OK, wow. Doing some checking around on files that have been unchanged since the first commits to the HighwayData repo, I'm only seeing `0A` newlines. Looks like `0D0A` was not as pervasive, since the beginning, as I thought, and problems may have been more isolated.
Also, there was [this commit](https://github.com/TravelMapping/HighwayData/commit/1307317419151a03145212bc6817138110cd7eee), following from some discussion [here](https://github.com/TravelMapping/HighwayData/pull/1072#issuecomment-274592924).

I noted in that discussion that the DIFFs looked funny. Jim merged the data in, and later made the commit to remove the CRLFs. But... did it actually cause any problems with siteupdate.py?

In any case, the new usaal WPTs have CRLFs in them, and siteupdate.py on FreeBSD seems to handle them just fine.
Maybe what we're seeing here is a difference in how (Python) things work on FreeBSD vs macOS...

@jteresco, do you remember any if these events/discussion; anything to add here?

It may be best to hold off on pulling this in, until we can do some tests on both macOS and Unix-like systems, to make sure we get the desired results. Optimally siteupdate.py should be able to successfully parse `0A`, `0D0A`, and `0D` as newlines, on any OS.
The `line.rstrip('\n')` that I'm changing here has been around since [the first commit](https://github.com/TravelMapping/DataProcessing/blob/078e1343bdaa082d304d3ceb181fc9c4f684f176/siteupdate/python-teresco/read_data.py) of what was then known as read_data.py. But it may not have been enough...
I'm not too familiar with how Python handles the `\n` escape sequence under the hood. If it denotes just `0A`, or works with `0D0A`, etc. If Python also supports `\r`, that may something to look into.

@jteresco, if you try out either the proposed or existing siteupdate.py on your Mac again, pay particular attention to what happens with whopperman.list. Also, how does it cope with the usaal WPT files?

michihdeu:
> Maybe what we're seeing here is a difference in how (Python) things work on FreeBSD vs macOS...

I think that's the issue.

When I download any GitHub repository to by Win PC, all files have CRLF. I've added a shortcut to Notepad++, I just need to press CTRL+u to convert the file. It's an automated manual process for me but it would be great if I could get rid of it one day...

jteresco:
Yes, in the "old days" I used to run siteupdate.py on my Mac and transfer to tm.teresco.org, which was just a pointer to blizzard.teresco.org.  Problems would come up when I occasionally would run siteupdate.py on my home FreeBSD machine for one reason or another.

Now that the whole mess is done on noreaster, I rarely run an update on my Mac any longer.  But I think it's worth me merging this in and pulling it to my Mac to make sure things are good there.  I assume you've tested on noreaster enough to be confident that it doesn't break things there.  Seem OK?

yakra:
@jteresco wrote:
> Problems would come up when I occasionally would run siteupdate.py on my home FreeBSD machine for one reason or another.

Is this something you've tried recently? Not in a long time?
Did problems come up on *just* the FreeBSD machine, or FreeBSD *and* the Mac?

@jteresco wrote:
>  I assume you've tested on noreaster enough to be confident that it doesn't break things there. Seem OK?

My tests on noreaster were limited to time trials. I had the program `sys.exit()` after reading in all the WPT data. The only changes were to `read_wpt` and the Waypoint ctor, and thus included in this, and I saw no error messages or crashes.

@yakra wrote:
> * In any case, I achieved the desired effect: in addition to newline-only lines, whitespace-only lines are not passed to the Waypoint constructor or parsed at all. DIFFs of [deterministic logs](https://github.com/yakra/DataProcessing/issues/17) affected only the creation time. Siteupdate.log reported the same number of total waypoints & unique locations. I observed no unintended side effects.

I did this on BiggaTomato, my Ubuntu 16.04 machine.
I don't believe I broke anything not already potentially broken.
The only differences in Python between FreeBSD & Ubuntu *that I'm aware of* are how they handle iterating thru some container formats, E.G. dicts and/or sets.
Shall we just say Unix-like is Unix-like, and that we're probably OK?
Seems OK on BiggaTomato.

@yakra wrote:
> The `line.rstrip('\n')` that I'm changing here has been around since [the first commit](https://github.com/TravelMapping/DataProcessing/blob/078e1343bdaa082d304d3ceb181fc9c4f684f176/siteupdate/python-teresco/read_data.py) of what was then known as read_data.py. But it may not have been enough...

I don't think we have anything to fear; no `line` should have any extraneous newlines at the end that should need to be stripped out. We know this because whopperman.list is parsed successfully. And all that .list file processing does is a `line = line.strip()`.

@jteresco wrote:
>  But I think it's worth me merging this in and pulling it to my Mac to make sure things are good there.

IMO a robust siteupdate that can run on a variety of systems is a Good Thing.
Again, I don't believe I've newly broken anything. What I'm interested in is, how does the existing siteupdate fare? On the Mac or on your home FreeBSD machine?

@michihdeu wrote:
> It's an automated manual process for me but it would be great if I could get rid of it one day...

I think you can safely get rid of it. (As long as we're running updates directly on noreaster, at least?)
usaal & all the other files Jim mentions [here](https://github.com/TravelMapping/HighwayData/issues/2439#issuecomment-449515879) work OK.

jteresco:
In the days before noreaster, the problems I would encounter were on FreeBSD being more picky about things than my Mac.  Now that everything runs on FreeBSD all the time, I don't expect we'll have problems.  I'll be merging in these changes and will try them first on my Mac, then if all goes well, I'll pull to noreaster for the next production site update.

I agree we should do our best to keep and improve robustness.  It's good that we regularly use these tools on FreeBSD and Ubuntu Linux, and occasionally still on Mac OS X's flavor of Unix.

jteresco:
These changes just ran successfully, best I can tell, through a datacheck.sh on my Mac, and through a full site update on noreaster.

yakra:
Testing out some changes for #88, I'm able to answer some questions I asked in the OP:
> * When we do a `lines = file.readlines()`, the `\n` delimiter is removed, and not contained in the lines in `lines`, correct?

Looks like this is not the case, and `\n` stays until stripped.

> * If `\n` is contained in the individual lines, does strip() get rid of it?

Yes. (So, it looks like there's nothing to fear.)

Reading system codes directly as lines from systemgraphs.csv did not work.
`strip()`ping the lines worked.

PR #163, opened 2019-01-07
move fatal error reporting before graph generation

### LINKED ISSUE #162: Failure in siteupdate.py reported much later
Recent changes have meant that when a data error slips in, the "ABORTING due to N errors" doesn't happen until after (the time consuming) generation of all the subgraphs.  I had gotten used to being able to ignore the site update process once graph generation started.  I'd like to see about getting that ABORTING condition moved back earlier so I have a better chance to notice the problem.  I just about was calling it a night when I noticed tonight's update had failed.  I'll try to check this out soon but wouldn't be sad if @yakra beat me to it.

yakra:
This closes #162.

Looks like we never had a `file.close()` for datacheckfps.csv.
I didn't add one, Shall we do so?

### DIFF
--- a/siteupdate/python-teresco/siteupdate.py
+++ b/siteupdate/python-teresco/siteupdate.py
@@ -2860,6 +2860,34 @@ for h in highway_systems:
     sysfile.write('\n')
     sysfile.close()
 
+# read in the datacheck false positives list
+print(et.et() + "Reading datacheckfps.csv.",flush=True)
+with open(args.highwaydatapath+"/datacheckfps.csv", "rt",encoding='utf-8') as file:
+    lines = file.readlines()
+
+lines.pop(0)  # ignore header line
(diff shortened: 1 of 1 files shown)


jteresco:
> Looks like we never had a `file.close()` for datacheckfps.csv.
> I didn't add one, Shall we do so?

Yes, please do.

yakra:
This is the case for updates.csv, systemupdates.csv, areagraphs.csv, systemgraphs.csv, multisystem.csv, multiregion.csv, datacheckfps.csv.
Shall I add `file.close()` for all of these?

jteresco:
Yes, I think that would be good.  It's not really hurting anything but it's good practice.

yakra:
Done.

jteresco:
Excellent, thanks.  Merging.

Issue #165, opened 2019-01-10
single quotes in groupname field


yakra:
In the ConnectedRoute constructor:
https://github.com/TravelMapping/DataProcessing/blob/5db8dfa9366dcc34bc7862041cd80fc3d86513b3/siteupdate/python-teresco/siteupdate.py#L881
This appears to be there to play nice with inserting the field into the SQL file. I presume `'` is an escape character.

However, there are unintended side effects in [highwaydatastats.log](http://travelmapping.net/logs/highwaydatastats.log):
`TunEnv (Tunel d''Envarlira): 2.55 mi (AND TunEnv only)`, rather than
`TunEnv (Tunel d'Envarlira): 2.55 mi (AND TunEnv only)`

Probably better to perform this replacement *when* `groupname` is written to the SQL file, rather than before.

jteresco:
I'm OK with changing it to do so only when writing the SQL, but I'm also not bothered much by the extraneous `'` in the log file.

yakra:
Also, when calling the HighwaySystem constructor:
https://github.com/TravelMapping/DataProcessing/blob/bc6eacdc5a1cbbe7c821ec43bc9b93d1539eeeaa/siteupdate/python-teresco/siteupdate.py#L2025
From a housekeeping perspective it seems sensible to do this replacement when writing to the SQL file too.
This doesn't matter very much; I don't believe this field is used in any of the .log files.
But if we ever wanted to enhance the user list logs, say, from
`System canon (active) overall: 4554.78 of 6285.55 mi (72.5%)` to
`System canon, Ontario King's Highways (active) overall: 4554.78 of 6285.55 mi (72.5%)`
this would be something to watch out for.

Just making note of this; I don't consider it important enough to open a new issue or reopen this one.

PR #166, opened 2019-01-10
stats log enhancements

yakra:


yakra:
859d92bc704d625fd8e90312cc0ac1281927ae73 closes #165.
I don't have means of testing it out, but this should work OK, right?

This is the only occurrence of `groupname` requiring double quotes.
The other two occurrences are for stats log generation, and should use single quotes instead.

Issue #133, opened 2018-11-15
nearmatchfps.log formatting


yakra:
```
DCERROR: chngs.g2012;['NX/GS', 'S201'];VISIBLE_DISTANCE;17.94
FPENTRY: chngs.g2012;NX/GS;S201;;VISIBLE_DISTANCE;18.61
REPLACEWITH: chngs.g2012;NX/GS;S201;;VISIBLE_DISTANCE;17.94
```
The `DCERROR` datum is redundant, just the same info as the `REPLACEWITH` datum formatted differently.
It makes sense to someone who's been digging into siteupdate.py & learning about Python, but is probably just extra noise for those used to seeing FP Entries formatted as they are on [datacheck.php](http://travelmapping.net/devel/datacheck.php).
I propose removing `DCERROR`.

The remaining info will be easier on the eyes if aligned a constant amount from the left.
Suggest `FPENTRY` -> `FP_ENTRY` and `REPLACEWITH` -> `CHANGETO`, yielding:
```
FP_ENTRY: chngs.g2012;NX/GS;S201;;VISIBLE_DISTANCE;18.61
CHANGETO: chngs.g2012;NX/GS;S201;;VISIBLE_DISTANCE;17.94
FP_ENTRY: chngs.g22;G70_E;G312_Luo;;VISIBLE_DISTANCE;16.72
CHANGETO: chngs.g22;G70_E;G312_Luo;;VISIBLE_DISTANCE;16.91
FP_ENTRY: chngs.g3011tua;G215;X280;;VISIBLE_DISTANCE;23.24
CHANGETO: chngs.g3011tua;G215;X280;;VISIBLE_DISTANCE;13.84
FP_ENTRY: chngs.g3017;X184;JinAir;;VISIBLE_DISTANCE;26.69
CHANGETO: chngs.g3017;X184;JinAir;;VISIBLE_DISTANCE;27.07
FP_ENTRY: chngs.g7011;SN/GS;Xia;;VISIBLE_DISTANCE;11.20
CHANGETO: chngs.g7011;SN/GS;Xia;;VISIBLE_DISTANCE;11.22
...
```

Since I've been getting my hands dirty with siteupdate.py a lot lately, I volunteer to make these changes if there's consensus it's a Good Thing.

jteresco:
Go for it!

michihdeu:
Cool, I finally understand what this log means! Yes, go for it!

PR #175, opened 2019-01-21
revised str(DatacheckEntry) based on CSV line

yakra:
str(DatacheckEntry) is only used when [sorting datacheck entries](https://github.com/TravelMapping/DataProcessing/commit/73107c87e04f2979a6cc9bb03dfd18704deb3b85) before marking false positives.

Before that, it was only used [in nearmatchfps.log](https://github.com/TravelMapping/DataProcessing/commit/12f00bd9ef3378394dff9de74b8e7de10c345932), but was [eliminated as redundant and distracting](https://github.com/TravelMapping/DataProcessing/issues/133).

Sorting asciibetically by the CSV line as we see it is more intuitive than sorting by a "Python list" style string we don't actually see in the log.
This would bring siteupdate.py more in line with the in-dev siteupdate.cpp, and have their datacheck.log files sorted in the same order, for more usable DIFFs.

@jteresco, is there any reason to keep the old "Python list" style str(DatacheckEntry) around?

### DIFF
--- a/siteupdate/python-teresco/siteupdate.py
+++ b/siteupdate/python-teresco/siteupdate.py
@@ -1171,7 +1171,17 @@ class DatacheckEntry:
         return True
 
     def __str__(self):
-        return self.route.root + ";" + str(self.labels) + ";" + self.code + ";" + self.info
+        entry = str(self.route.root)+";"
+        if len(self.labels) == 0:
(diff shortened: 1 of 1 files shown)


jteresco:
No need to keep.  We can always make one again if it's needed for something later.

PR #188, opened 2019-02-05
graph construction cleanup

yakra:
`self.vertex1` & `self.vertex2` were already defined as `graph.vertices[s.waypoint1.unique_name]` & `graph.vertices[s.waypoint2.unique_name]`, making these lookups unnecessary. Saves about 0.4s during graph setup.

The `HighwayGraphCollapsedEdgeInfo` constructor is already done this way.

### DIFF
--- a/siteupdate/python-teresco/siteupdate.py
+++ b/siteupdate/python-teresco/siteupdate.py
@@ -1256,17 +1256,17 @@ class HighwayGraphEdgeInfo:
         # checks for the very unusual cases where an edge ends up
         # in the system as itself and its "reverse"
         duplicate = False
-        for e in graph.vertices[s.waypoint1.unique_name].incident_edges:
+        for e in self.vertex1.incident_edges:
             if e.vertex1 == self.vertex2 and e.vertex2 == self.vertex1:
                 duplicate = True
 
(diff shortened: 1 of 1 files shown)


yakra:
May have some more commits to add to this.

yakra:
31c960223ccc1f0c033067ed885b5c41e85bf7b5:

We can visit each concurrrent HighwaySegment only once by checking whether it's the first segment in its concurrency list. Doing so removes the need to check a name for each to see if it's been visited, thus removing the need to store the `visited` & `segment_name` variables for each HighwaySegment. Thus we can remove 2 FOR loops (29 lines of code) and construct edge rights away, getting segment names from `HighwaySegment.segment_name()` during construction.

Calling the `HighwayGraphEdgeInfo` and initial `HighwayGraphCollapsedEdgeInfo` constructors have been combined into the same FOR loop.

yakra:

yakra:
This one should be OK.

jteresco:
I'm holding off on merging in any additional graph generation changes while I'm using the graphs heavily in class.  One more week of that then we'll have a break from all that and I'll be a little more adventurous.

yakra:
Understandable. That graphs are used in your classes is a pretty important reason to make sure any changes are done right, and that things behave in an expected, predictable way.

PR #190, opened 2019-02-07
Unique names cleanup

yakra:
This branch also contains the commits in #188.

Right now, vertex labels are stored in four places:
* as keys of `unique_waypoints` dict
* within waypoints
* within vertices themselves
* as keys of `vertices` dict

With 808,995 points at 551,932 unique locations, the memory usage can add up. 
With a slightly older revision of the HighwayData repo, I calculated potential savings of 24,703,931 B for just the strings themselves.
Add in the overhead of the std::string container itself, and it goes up to 86,716,079 B savings in C++.
(I don't know what the string container overhead works out to in Python.)

This cuts down on the duplication:
* The `unique_waypoints` dict is replaced with a set, to track the unique names as they're generated.
  - Unique waypoint locations themselves are determined by whether a waypoint is a singleton waypoint, or at the front of its colocation list.
  - This set is deleted when vertex construction is complete and it's no longer needed.
* No more storage within waypoints.
  - The HighwayGraph constructor instead passes `point_name` directly to the vertex constructor, cutting out the Waypoint object middleman.
* The vertices themselves are now the only place permanently storing unique names.
* The `vertices` dict uses canonical waypoint objects as keys, rather than unique names.

This also results in slight time savings during graph setup, about 3.8s on BiggaTomato.

### DIFF
--- a/siteupdate/python-teresco/siteupdate.py
+++ b/siteupdate/python-teresco/siteupdate.py
@@ -1256,17 +1256,17 @@ class HighwayGraphEdgeInfo:
         # checks for the very unusual cases where an edge ends up
         # in the system as itself and its "reverse"
         duplicate = False
-        for e in graph.vertices[s.waypoint1.unique_name].incident_edges:
+        for e in self.vertex1.incident_edges:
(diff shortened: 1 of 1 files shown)


yakra:

PR #191, opened 2019-02-08
HighwayGraphCollapsedEdgeInfo initial setup

yakra:
This branch also contains the commits in #188 & #190.

With the initial HighwayGraphCollapsedEdgeInfo setup based on a HighwaySegment, we end up recomputing a lot of the info we just obtained for the HighwayGraphEdgeInfo based on the same HighwaySegment. If we base initial construction on the corresponding HighwayGraphEdgeInfo instead, we can avoid duplicating this work.
* segment_name is copied from the HighwayGraphEdgeInfo, rather than recomputed from HighwaySegment.segment_name().
* vertex1 & vertex2 are copied from the HighwayGraphEdgeInfo, avoiding looking them up by Waypoint in the hash table.
* "Reverse" duplicate checks are done away with, as we already know the HighwayGraphEdgeInfo to not be a duplicate.
* If a HighwayGraphEdgeInfo is a duplicate, we skip its corresponding HighwayGraphCollapsedEdgeInfo entirely.

Compared to the version in #190, this reduces initial HighwayGraph setup time from ~41.8s to ~38.5s on BiggaTomato.

---

~**ToDo:** The HighwayGraphCollapsedEdgeInfo constructor receives a HighwayGraph object, but doesn't do anything with it. This can be trimmed.~ **Done:** 268485b0d4758496a331528c72ec5645291e21cf

### DIFF
--- a/siteupdate/python-teresco/siteupdate.py
+++ b/siteupdate/python-teresco/siteupdate.py
@@ -1267,7 +1267,6 @@ class HighwayGraphEdgeInfo:
         if not duplicate:
             graph.vertices[s.waypoint1.unique_name].incident_edges.append(self)
             graph.vertices[s.waypoint2.unique_name].incident_edges.append(self)
-            graph.edges.add(self)
 
     # compute an edge label, optionally resticted by systems
     def label(self,systems=None):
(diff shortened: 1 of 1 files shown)


yakra:

Issue #192, opened 2019-02-09
PlaceRadius.contains_edge


yakra:
Our list of `matching_vertices` is already known.
Rather than perform all the trig functions and compare distance to the radius, it *may* be faster to check both points for membership in the list.

yakra:
The changes for #187 will change `matching_vertices` from a list to a set. This is a Good Thing, as checking for membership in a set is faster then a list.

yakra:
C++: Area graph generation took 12.8s, compared to 12.6 [before](https://github.com/TravelMapping/DataProcessing/issues/187#issuecomment-462162815).
Still worth giving this a try in Python; it's a simple enough modification.

yakra:
Python: Area graph generation took 49.5s, compared to 50.2 [before](https://github.com/TravelMapping/DataProcessing/issues/187#issuecomment-462236303). This is either a small decrease in processing time, or within margin of error.
I'll run some more tests to try to get a better handle on which is the case.

yakra:
Looks like we're faster by about half a second:
```
math		50.8	50.6	50.4	50.3	50.2
membership	50.8	50.3	49.9	49.5	49.3
```
Another idea that may be faster than checking set membership:
When adding vertices to a new per-PlaceRadius set, use the `vertex_num` variable to flag it as part of the set. Check `vertex_num` when `matching_edges`.

yakra:
```
Area Graphs
====================================================
math		50.8	50.6	50.4	50.3	50.2
vertex_num	49.3	48.7	48.5	47.8	47.2


Setup
====================================================
math		26.7	26.2	26.1	25.9	25.8
vertex_num	27.8	27.4	27.4	26.6	26.3
```
Heh. So we're saving 0.9-3.4s, 2.2s on average.
And possibly taking up about 1s more during setup, due to initializing two lousy variables here, though this is getting to be within margin of error. For the sake of Jim's sanity, I'm not going to bother opening a pull request right away, but here's the code: https://github.com/yakra/DataProcessing/commit/622da4f885fb1099a3e0f4ce2eb8580ee6b1c5d8

yakra:
Back to C++ now, looking at the `vertex_num = -1` method...
```
Setup
========================================================
math		11.9	11.9	11.9	11.9	11.8
vertex_num	11.9	11.9	11.9	11.9	11.9

Area Graphs
========================================================
math		13.0	12.8	12.8	12.8	12.8
[...]

PR #193, opened 2019-02-11
per-region/system vertex sets

### LINKED ISSUE #187: sets & unions for vertex/edge matching
Another idea to look into...

For each system, store a set of
* vertices
* edges
* collapsed edges

Regions are trickier; in siteupdate.py there's no Region class, only tuples of strings.
Short of rewriting large swaths of the code, store these same items in a dictionary.

yakra:
This branch also contains the commits in #188, #190, & #191.
Closes #187.

https://github.com/TravelMapping/DataProcessing/blob/f07a843abe1719a6411e4d62c33a6e55a69552fb/siteupdate/python-teresco/siteupdate.py#L1674-L1681
wasn't strictly necessary, but I added it to maintain the full functionality of the old code.
While these cases don't show up in the real world with the types of graphs we're making now, we nonetheless maintain the ability to write graphs restricted by PlaceRadius *and* region/system. (So if we ever want a graph of usai & usaus in MD & VA within the DC PlaceRadius...)
An extra `if` or two won't hurt much; this is already plenty fast. :)

I'll have a go at #192; after that I don't plan any more big changes to graph generation. ...In Python, at least. ;)

### DIFF
--- a/siteupdate/python-teresco/siteupdate.py
+++ b/siteupdate/python-teresco/siteupdate.py
@@ -1267,7 +1267,6 @@ class HighwayGraphEdgeInfo:
         if not duplicate:
             graph.vertices[s.waypoint1.unique_name].incident_edges.append(self)
             graph.vertices[s.waypoint2.unique_name].incident_edges.append(self)
-            graph.edges.add(self)
 
     # compute an edge label, optionally resticted by systems
     def label(self,systems=None):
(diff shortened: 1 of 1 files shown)


jteresco:
I just ran what appears to be a successful "practice" site update on my Mac, and the speedup is fantastic!

PR #198, opened 2019-03-22
graph setup progress bar

### LINKED ISSUE #194: Progress bar output for setting up graphs
It would be nice to have a progress bar here. Nothing "happens" for about 30s when executing `datacheck.sh`.

```
[283.1] Setting up for graphs of highway data.
Full graph has 550494 vertices, 622495 edges.
Edge compressed graph has 399056 vertices, 471057 edges.
[314.5] Writing graph waypoint simplification log.

```

yakra:
This closes #194

### DIFF
--- a/siteupdate/python-teresco/.gitignore
+++ b/siteupdate/python-teresco/.gitignore
@@ -1,3 +1,5 @@
 graphs
 logs
 nmp_merged
+stats
+TravelMapping.sql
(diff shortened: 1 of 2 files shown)


yakra:
**.gitignore:**
stats/ is typically produced by datacheck.sh, but TravelMapping.sql is not. If preferred, I can remove `TravelMapping.sql` from .gitignore, leaving it to the individual collaborators to add it to  their .git/info/exclude files.

jteresco:
I don't think extra entries in .gitignore cause any harm, so I would prefer to have it in there.

PR #224, opened 2019-05-19
great-circle distance speedups

yakra:
### C++:  store & retrieve segment length
Increases speed slightly when Computing stats & Creating per-traveler stats logs.
Along for the ride: added alternate great-circle distance formulas to Waypoint::distance_to & PlaceRadius::contains_vertex, commented out for now

### DIFF
--- a/siteupdate/cplusplus/classes/GraphGeneration/HGEdge.cpp
+++ b/siteupdate/cplusplus/classes/GraphGeneration/HGEdge.cpp
@@ -49,9 +49,7 @@ HGEdge::HGEdge(HighwaySegment *s, HighwayGraph *graph)
 }
 
 HGEdge::HGEdge(HGVertex *vertex, unsigned char fmt_mask)
-{	// build by collapsing two existing edges around a common
-	// hidden vertex waypoint, whose information is given in
-	// vertex
+{	// build by collapsing two existing edges around a common hidden vertex
 	c_written = 0;
 	t_written = 0;
 	format = fmt_mask;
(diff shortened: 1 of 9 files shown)


yakra:
Added a commit to implement this in siteupdate.py.

### On noreaster:
* *Reading waypoints for all routes* 116.9 -> 120.1 s, gaining 3.2 s.
* *Computing stats* 30.1 -> 26.1 s, losing 4.0 s.
Let's call this moving 3.6 s +/- 0.4 s from point A to point B. Times always seem to vary a little bit between runs in Python *(for example, the total run time for all tasks other than the ones I measured for this benchmark went 260.2 -> 254.8 s)*; there will be a little noise in the signal.
* *Creating per-traveler stats logs and augmenting data structure* 46.2 -> 29.9 s. A 50% speedup for this task.

yakra:
### Spherical law of cosines formula for area graphs
253f75270dfee3650d05a5962e44074e7284f5a2 http://forum.travelmapping.net/index.php?topic=3014.msg13996#msg13996

Same as the existing formula, but with some terms factored out or simplified via trig identity.
Increases area graph generation speed by ~32%.
* Not such a big deal in C++ on BiggaTomato, where it gets us from 3.7 to 2.8 s.
* In Python on noreaster, we go from 56.8 to 43.1.

No DIFFs in the resulting graphs.

jteresco:
Sorry I forgot this was here.   Pulling in now and it will be used starting with tomorrow's site update.  And thanks again for all the work on this.

PR #227, opened 2019-06-19
quadtree-based area graph vertex search

### LINKED ISSUE #223: area graph vertex matching: quadtree
Creating each area graph, we measure the distance from the area's center to *every* vertex in the graph structure, and compare against the radius.

What if we measured the distance of only a subset of vertices?
* We have a center and a radius, so we should be able to get latitudes/longitudes for a bounding box within which to search.
  - ToDo: make sure min/max longitudes remain sane as we approach the poles.
* Recurse down through the WaypointQuadtree almost *(but not quite)* exactly the same as we do when finding near_miss_waypoints.
* Each Waypoint found gives us an index to the `HighwayGraph.vertices` dict. Do whatever's fastest, either:
  - Construct a set of vertices out of the quadtree contents,
    - check if singleton or at front of colocation list, and if so add to `vertex_set`, or
    - add directly to `vertex_set` by `hashpoint()`, letting the set() methods determine whether the `hashpoint()` has already been added.

    ...then measure distances to each vertex in this subset, removing those outside the radius. Or make a new set including those within it, whichever we need to do.
  - Or alternately, just check if singleton or at front of colocation list, and within radius, *then* add to `vertex_set`. Fewer additions and thus membership checks, and no need to re-iterate thru the set to remove outliers.

### LINKED ISSUE #226: C++: traveled subgraph DB entries
\# number of vertices & edges are garbage: uninitialized variables
\# of travelers = the total number of TravelerLists, not the # of travelers clinching @ least 1 segment on the subgraph

This will be fixed in my solution for #223.

yakra:
This closes #223.
This closes #226.


yakra:
Ran fine on BiggaTomato. Just crashed on noreaster. Going to go sort this out.

yakra:
Ready to rock.

jteresco:
Thanks - it will be used for tonight's update.

PR #259, opened 2019-11-16
ev_combine

### LINKED ISSUE #233: single matching_edges function
With separate `matching_simple_edges`, `matching_collapsed_edges` and `matching_traveled_edges`, we iterate thru the entire set of vertices & check for matches 3 times.
Combine into one function, and iterate only once, match multiple edge types at the same time.

yakra:
Tidied up subgraph-generation code by matching all vertices and simple, collapsed & traveled edges in a single function, yielding a very small speedup. Closes #233.

**Python:** Saves ~4s on BiggaTomato. We're down 23 lines of code.
**C++:** Saves only ~1s, with more substantial code-cleanup benefits here; we're down 76 total lines across several files. In the process, we save some RAM by getting rid of the underwhelming per-Region & per-HighwaySystem edge sets from #187, no worse for wear performance-wise.


yakra:
Due diligence: I will test the Python version on noreaster, then reopen.

yakra:
Looks good. Ready to rock.

PR #270, opened 2019-12-12
magical 1-byte speed increase

yakra:
Creating & destroying an array of 6 strings can add up over the course of some 19,630,920 iterations. Passing the datacheck FP entries by reference instead of value avoids this, yielding a substantial speed boost. Seems the bottleneck here is memory bandwidth; the more bandwidth my machines have, the more muted the effect of the speedup.

* BiggaTomato (DDR2-800): 10.3s -> 2.6s, a 3.96x speedup.
* Epoch (DDR3-1333): 10.7s -> 4.1s, a 2.61x speedup.
* lab2 (DDR3-1600): 4.1s -> 1.9s, a 2.16x speedup.
* noreaster is the exception: 6.7s -> 1.3s ~= 5.15x. I'll just assume that noreaster has a good number of other processes running and competing for bandwidth. (I *assume* noreaster has the highest bandwidth. It could be running at <1600 mhz, or even single-channel for all I know...)


jteresco:
Just missed by a few weeks being in time to show this as an example of differences in calling conventions in my programming languages class.

yakra:
> Seems the bottleneck here may be memory bandwidth

Or not. I installed a single stick of DDR3-1333 in lab2. 4.2s -> 2.2s. So there are other factors at play here. Cache size, maybe? Epoch has 512K, BiggaTomato 8MB, and lab2 has 512K L2 + 20M L3. And if noreaster has a bunch of other processes running, that could mean more data in the cache for them, and more cache misses for siteupdate. In any case, the  speedup is real, and it's a Good Thing.

> Just missed by a few weeks being in time to show this as an example of differences in calling conventions in my programming languages class.

:D
I've scoured the code looking for opportunities for similar improvements. There are a few, but nothing of this much benefit, that gets repeated this many times.
Other possibilities to check:
* range `for` loops
* return values

jteresco:
Noreaster has all the web service and DB service, which I suppose could be using some memory bandwidth, but I wouldn't expect that much.  No CPU intensive or other memory intensive jobs are normally running at this point.  The machine's been up for a while so the bootup logs aren't easy to see at the moment to remind myself of its memory size and type and cache sizes.

In any case, a nice speedup!

Issue #271, opened 2019-12-14
Unused alt labels sorted per region


michihdeu:
http://travelmapping.net/logs/unusedaltlabels.log

The file is huge and not sorted by region. It's hard for a highway data manager to find his relevant entries.

It would be nice to have a file per region like we have for [NMPs](http://travelmapping.net/logs/nmpbyregion/).

yakra:
Or how about the same single file, just sorted by region?

yakra:
Until the file actually comes pre-sorted, on noreaster you can type
`sort unusedaltlabels.log | less`
to sort the file and then pipe it into a basic viewer that lets you scroll around with arrow keys, Page Up, Page Down, Home and End.

michihdeu:
I just want the file being empty........ Like data check......

I'm able to find the relevant lines but I'm not sure that all highway data managers find their lines nor that they are even aware of it....................... it's just.....

jteresco:
To me, worrying about the unused alt labels is a pretty low priority.  There's little cost to the existing at site update time, and I think zero cost in the DB or front end.  The only time they matter is when a waypoint label is added or changed and it conflicts with an existing unused alt label.

If we wanted to clean them all up, an automated process seems to be in order.  One pass and we're done, then can watch for new ones creeping in as highway data and list files change.

yakra:
@jteresco wrote:
> There's little cost to the existing at site update time,

The only cost is when a label in a .list file line doesn't match a point's primary label. Then, we iterate thru AltLabels. So the only cost here is when unused AltLabels come before used ones in a given point's AltLabel list, and we have to interate past & check the unused ones before matching the used ones. Not a big cost.

@jteresco wrote:
> and I think zero cost in the DB or front end

Yep. AltLabels are handled in the siteupdate side, during the `TravelerList` constructor. None of that info ever makes it into the DB.

@jteresco wrote:
> If we wanted to clean them all up, an automated process seems to be in order. One pass and we're done, then can watch for new ones creeping in as highway data and list files change.

For my part, I've been purposefully leaving a few in:
* PE1, AB1, AB16, MB1 and MB100 all have unused AltLabels, which are still in use in their corresponding TCH routes. I've only eliminated AltLabels unused in *both* routes, to ensure there are no diffs between the corresponding TCH / provincial files.
* Leaving notes:
[...]

PR #273, opened 2019-12-16
sort unusedaltlabels.log lines

yakra:
For #271. Not a file per region as @michihdeu originally suggested, but instead keeping the original single file, only sorted by region.


jteresco:
Good to go live tonight?

yakra:
Good to go live.

jteresco:
Pulled onto noreaster for the site update in progress, thanks.

PR #280, opened 2019-12-30
no-mutex NMPs

### LINKED ISSUE #141: idea: replace per-point NMP lists with a method
Replacing per-point NMP lists with a method could potentially save us a bit of RAM headroom.
Overall CPU time would be about a wash, I think.
Something to investigate.

yakra:
Closes #141.
More improvements to come, that should also benefit the Python version.


yakra:
> More improvements to come, that should also benefit the Python version.

The code has been committed since Dec 30th. I'm working on making some pretty graphs to go along with the pull request.

PR #284, opened 2020-03-21
improved speed & parallelism when reading WPTs

yakra:
Closes https://github.com/yakra/DataProcessing/issues/111. This picks up where #280 left off, improving multi-threaded performance `Reading waypoints for all routes` by moving unique location counting & colocation detection *(parts of these improvements also found their way into siteupdate.py, removing some redundant iteration down thru quadtree nodes)* to where it's not necessary to keep the quadtree locked, and then finally reworking WaypointQuadtree::mtx to lock individual nodes, allowing different branches of the quadtree to be refined in parallel.

Graphs were produced by grepping the beginning and ending timestamps for a given task in siteupdate.log files (3000 in total!) produced by the different binaries, averaging the times of 10 different runs each.

* The blue line shows time `reading waypoints for all routes` before #280, back when NMP detection was still included in this task.
* The 3 red-orange lines represent the changes in #280, when NMP detection was carved out to be done separately. The solid line, the sum of the two dotted/dashed lines, is the time spent performing both tasks, and thus our basis of comparison against the blue line.
* The remaining (yellow, green, purple) lines are for reading WPTs only, sans NMPs, and should be compared to the upper dashed line, "readWPT".

---
lab2
IMO our sanest & nicest-looking graph overall. Lab2's 20MB L3 cache helps it scale up to a higher number of threads, making it just barely faster at this task than the faster-clocked lab1.

---
lab3
Here we see our bottleneck(s) as a big factor in 30d84fd, and the effects of removing it.
I would *assume* that if our only bottleneck were the mutex, with nothing else at play, our time would make an asymptotic approach to the time required to do the effectively single-threaded bits, and then stay relatively flat. What we see is efficiency peaking early, and things quickly getting worse. My suspicion = memory; lab3 has 15MB L3 cache, and 1333 MHz RAM vice lab2's 1600. When we have to start going (earlier) to slower main memory for our data, it's all over.
By contrast, 10f9c92 calms things down quite a bit. Once NMPs are out of the equation, we're working with a smaller data set at once, and can keep more of what we're using in cache.

---
lab1
Raw clock speed gives lab1 an initial edge, but it starts to feel the slowdown a bit earlier. 8MB L3 cache here.

---
lab1_5
Wow. Worst of both worlds on lab1.5: 8MB L3 cache + 1333 MHz RAM clock = ouch!
What we ***don't*** see here: While expectedly slower at this particular task, lab1.5 runs siteupdate faster overall than lab1 (speed records are 42.5s & 48.5s respectively). This despite a Xeon E3-1230 CPU (vice an E3-1230 v2) with slightly slower clock speed and 1333 MHz RAM instead of 1600. I've not yet checked out what tasks account for the time savings.
Potential factors: Different OS. Different chipset?

---
bt
[And now last, but not to be the least,](https://www.youtube.com/watch?v=ppttxQoc74Q&t=1m32s) BiggaTomato, my daily driver desktop. The 30d84fd line looks a bit funky here, but it doesn't appear to be an artifact of low sample size; the data checks out. Weird.

---
If you'd like, I'll give an extra run thru the stats, logs and graphs and make sure everything checks out compared to what's in the master branch now.

### DIFF
--- a/siteupdate/cplusplus/classes/Route/read_wpt.cpp
+++ b/siteupdate/cplusplus/classes/Route/read_wpt.cpp
@@ -63,9 +63,7 @@ void Route::read_wpt
 			  if (al[c] >= 'a' && al[c] <= 'z') al[c] -= 32;
 			unused_alt_labels.insert(al);
 		}
-		all_waypoints->mtx.lock();
 		all_waypoints->insert(w, 1);
-		all_waypoints->mtx.unlock();
 
 		// single-point Datachecks, and HighwaySegment
 		w->out_of_bounds(datacheckerrors, fstr);
--- a/siteupdate/cplusplus/classes/Route/read_wpt.cpp
+++ b/siteupdate/cplusplus/classes/Route/read_wpt.cpp
@@ -63,20 +63,8 @@ void Route::read_wpt
 			  if (al[c] >= 'a' && al[c] <= 'z') al[c] -= 32;
 			unused_alt_labels.insert(al);
 		}
-		// look for colocated points
 		all_waypoints->mtx.lock();
-		Waypoint *other_w = all_waypoints->waypoint_at_same_point(w);
-		if (other_w)
-		{	// see if this is the first point colocated with other_w
-			if (!other_w->colocated)
-			{	other_w->colocated = new std::list<Waypoint*>;
-						     // deleted on termination of program
-				other_w->colocated->push_front(other_w);
-			}
-			other_w->colocated->push_front(w);
-			w->colocated = other_w->colocated;
-		}
-		all_waypoints->insert(w);
+		all_waypoints->insert(w, 1);
 		all_waypoints->mtx.unlock();
 
 		// single-point Datachecks, and HighwaySegment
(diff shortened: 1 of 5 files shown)


jteresco:
I'm willing to give the python version a shot with tonight's update.  Reasonable?

yakra:
Reasonable, yes, though I'd hate for any unintended glitches to sneak into the "official" commit history.
I was thinking more of the Python version when I made that comment.

yakra:
Gimme maybe 30 minutes? ;)

jteresco:
I'm going to give it a shot.

jteresco:
Sure.  Comment here or email when you're ready for me to launch a site update.

yakra:
Will comment when ready. Running both versions of siteupdate.py.
I have a shell script that automates comparing stats, logs, nmp_merged files, and graphs. Makes that all a lot easier.

yakra:
LOL nevermind! ;)
Noticed I was running siteupddate.py with one thread as per my SOP, decided that the "full abundance of caution route" would be to use 4 threads, and considered starting over...

yakra:
Ready!

jteresco:
And launched, thanks.

yakra:
Python DIFFs are good.

yakra:
C++ DIFFs good too.

Edited the OP to add:
>  Though lab3 is a dual CPU machine with a RAM bank for each CPU socket, this does not seem to double its memory bandwidth.

yakra:
ping @seaweedfactory @fl0gnap -- pretty pictures!

PR #305, opened 2020-04-12
error checks for field lengths

### LINKED ISSUE #56: Datachecks for field lengths
A system should be put in place where all entries from TM data files that end up in DB fields are checked for length to avoid problems like https://github.com/TravelMapping/Web/issues/195 .

A set of constants should be introduced to the site update program to make sure these values are consistent among all checks and when DB tables are created.

### LINKED ISSUE #247: error checks in C++ but not Python
**siteupdate.cpp**
```C++
el.add_error("Could not parse updates.csv line: [" + line + "], expected 5 fields, found 1");
el.add_error("Could not parse systemupdates.csv line: [" + line + "], expected 5 fields, found 1");
```
---
**Route.cpp** (the 2nd & 3rd tie into #56)
```C++
el.add_error("Unrecognized region in " + system->systemname + ".csv line: [" + line + "], " + line.substr(left+1, right-left-1));
el.add_error("Banner >6 characters in " + system->systemname + ".csv line: [" + line + "], " + banner);
el.add_error("Abbrev >3 characters in " + system->systemname + ".csv line: [" + line + "], " + abbrev);
```
---
**ConnectedRoute.cpp** (#56 again)
```C++
el.add_error("Banner >6 characters in " + system->systemname + "_con.csv line: [" + line + "], " + banner);
```
---
**HighwaySystem.cpp**
```C++
el.add_error("Could not find country matching " + systemsfile + " line: " + line);
```

### LINKED ISSUE #295: check all CSV lines for correct # of fields
to be addressed in tandem with #56 & #247

### LINKED ISSUE #297: PlaceRadius.place
The word "place" WRT area graphs is ambiguous. It can refer to:
* the [place](https://github.com/TravelMapping/DataProcessing/blob/889af2dba73d471cfe68c5cc33f76d745a692f61/siteupdate/python-teresco/siteupdate.py#L1603) property of a `PlaceRadius` object, corresponding to the 1st column of [areagraphs.csv](https://github.com/TravelMapping/HighwayData/blob/7482ea9e55271c0fa5f67544de18761c115dd883/graphs/areagraphs.csv#L1).
* the 2nd column of [areagraphs.csv](https://github.com/TravelMapping/HighwayData/blob/7482ea9e55271c0fa5f67544de18761c115dd883/graphs/areagraphs.csv#L1), corresponding to the [base](https://github.com/TravelMapping/DataProcessing/blob/889af2dba73d471cfe68c5cc33f76d745a692f61/siteupdate/python-teresco/siteupdate.py#L1604) property of a `PlaceRadius` object.

I'd like to clarify, and change the `PlaceRadius` variable to `descr`, and the areagraphs.csv column to base. Any objections?

### LINKED ISSUE #298: HighwaySystem ctor: deleted pointer / non-numeric Tier value
The HighwaySystem ctor assumes valid systems.csv input; it isn't designed to cope with a non-numeric Tier value:
`tier = *strtok(0, ";")-48;`
We dereference the pointer to the Tier field string to grab just the first character, and convert from a `char` to the int it represents.
So for a value of `QUACK!`, we grab the `Q`, ascii 81, and subtract 48 to arrive at a tier value of 33.
Back at the siteupdate `main` function, we call `hs->is_valid()` which returns False because it checks for a tier > 5.
So we delete the invalid `HighwaySystem`'s pointer -- and then attempt to follow it, resulting in reams of garbage output to the terminal.
https://github.com/TravelMapping/DataProcessing/blob/54a8482ddeb6128eb4f3e800690981c88e526da1/siteupdate/cplusplus/siteupdate.cpp#L207-L209
Change this to:
```C++
if (hs->is_valid())
{	highway_systems.push_back(hs); 
	cout << hs->systemname << '.' << std::flush;
}
else delete hs; 
```

Let's keep the check for a tier value of <1; that's simple enough. Changing `tier` from an `unsigned short` to a `short` will flag more values that are too low, instead of converting negative to large positive numbers.

Leaning toward scrapping the check for a tier > 5:
* The better way to do this would be to [use a constant](https://github.com/TravelMapping/DataProcessing/issues/56), and that seems a bit mission-creepy.
* The site could conceivably use larger tier #s in the future, and AFAIK larger values pose no problem on the Web side of things.

As far as doing a check for mixed alphanumeric values such as `3A` or full-on non-numeric, it comes down to how much we really care about doing this. **Edit:** Done.

* [x] fix siteupdate.cpp
* [x] `unsigned short` -> `short`
* [x] scrap the check for tier > 5
* [x] ~replace `HighwaySystem::is_valid()` function with a variable. Simpler, and could do a better job at invalid Levels starting with 'a', 'p', or 'd'.~ Nah. If we make it to the point where a Level value is assigned, we're good. Check for that.

yakra:
closes #56
closes #247
closes #295
closes #297
closes #298
closes https://github.com/yakra/DataProcessing/issues/120


yakra:
Ping @fmccann -- this is what happens when I refuse to eve give up 0.1s of speed:
https://github.com/TravelMapping/DataProcessing/blob/cb5270a5e92db6014463fd18c3bd05edb8897c48/siteupdate/cplusplus/functions/split.cpp#L1-L10

PR #307, opened 2020-04-14
malformed URL cleanup & robustness

### LINKED ISSUE #299: string -> number conversions
**Py:** We can avoid crashes when attempting to convert a non-numeric string to a number by using `try`/`except`/`else`.

Some (but not all) applications:
* [x] [Tier values in systems.csv](https://github.com/TravelMapping/DataProcessing/issues/298)
* [x] The MALFORMED_LAT & MALFORMED_LON datachecks could be rewritten to be more simple & compact. And *maybe* it might even help efficiency in the process.
* **C++:** malformed URL datachecks [already use](https://github.com/TravelMapping/DataProcessing/blob/166bc2c08340557c30d8bd92589fcd3ad4e370dd/siteupdate/cplusplus/classes/Waypoint/Waypoint.cpp#L112-L113) [strtod](http://www.cplusplus.com/reference/cstdlib/strtod/). ~A similar wasted opportunity for brevity and probably speed.~
* [x] search cplusplus/ for `strto`
* [x] search siteupdate.py for `int(` and `float(`

### LINKED ISSUE #302: MALFORMED_LAT, MALFORMED_LON & LABEL_TOO_LONG robustness
Need to take care to get the MALFORMED_LAT, MALFORMED_LON & LABEL_TOO_LONG datachecks right. Data inserted into `datacheckErrors.label1` and `datacheckErrors.value` must be neither:
* overlength
* truncated in the middle of a multi-byte character

It's very rare that this would actually happen, but gotta do it right.

---
LABEL_TOO_LONG cases should be [ignored, similar to the malformed URL cases](https://github.com/TravelMapping/DataProcessing/issues/57#issuecomment-449656914).
**Rationale:** To not ignore it means putting it in the DB, which means truncating the label to something of legal length. This truncated label could then end up in a .list file, and thus pointsinuse.log. Copying from there and pasting into wptedit would not flag the label as in-use, because the full label from the .wpt code doesn't match the truncated one from pointsinuse.log. It could too easily lead us to breaking .list files.

### LINKED ISSUE #306: Time for an @yakra addition as another author on this
https://github.com/TravelMapping/DataProcessing/blob/7253076524e54e49aff9ddf5ddffddab7d630226/siteupdate/python-teresco/siteupdate.py#L6

yakra:
closes #299
closes #302
closes #306

### DIFF
--- a/siteupdate/cplusplus/classes/Waypoint/Waypoint.cpp
+++ b/siteupdate/cplusplus/classes/Waypoint/Waypoint.cpp
@@ -44,74 +44,35 @@ Waypoint::Waypoint(char *line, Route *rte, std::mutex *strtok_mtx, DatacheckEntr
 		lat = 0;	lng = 0;	return;
 	}
 	bool valid_coords = 1;
-
-	// make sure lat= string is valid
-	size_t point_count = 0;
-	for (const char *c = URL.data()+latBeg; *c != '&' && *c != 0; c++)
-	{	// check for multiple decimal points
-		if (*c == '.')
-		{	point_count += 1;
-			if (point_count > 1)
-			{	//std::cout << ("\nWARNING: Malformed URL in " + route->root + ", line: ") + line << std::endl;
-				size_t ampersand = URL.find('&', latBeg);
-				datacheckerrors->add(route, label, "", "", "MALFORMED_LAT",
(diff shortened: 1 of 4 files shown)


jteresco:
Close and reopen mean we're good to go live with this for tonight?

yakra:
Good to go live. I considered and rejected making a cosmetic change to a couple lines.

Issue #281, opened 2020-02-28
threaded terminal output


yakra:
Different threads write to the terminal concurrently, resulting in output like
`MD (2805,3438) (2516,3149) (2516,3149) MP (190,237MOZ ) ((155,202) (1015155,1016) (,411202) ,412) (411,412) MTQ (339,459) (317,437) (317,437)`

**Solution:** Use `+` instead of `<<` operator

jteresco:
With +, are the writes atomic enough to avoid the overlap completely, or will it just improve the chances of avoid bad interleavings?  If we care a lot, all terminal output could be funneled back to the original thread for output.

yakra:
Atomic enough to avoid the overlap completely, I believe. I've switched to + before to solve this type problem.

`+` in this context is the string concatenation operator, so we first construct a std::string and then insert the whole thing into std::cout. Contrast multiple `<<` operators inserting multiple C strings / char arrays or std::strings in multiple operations.

yakra:
> If we care a lot, all terminal output could be funneled back to the original thread for output.

This may be what we need to get [Augmenting travelers for detected concurrent segments](https://github.com/yakra/DataProcessing/issues/117) to scale beyond [~3-5 threads](https://github.com/yakra/DataProcessing/issues/116#issuecomment-600934807).

yakra:
Experimentation for https://github.com/yakra/DataProcessing/issues/117 revealed this may not be as atomic as I thought. Sure, that uses an `ofstream` and this uses `cout`, but they're both forms of `ostream` to my understanding, and ought to behave similarly.
I might give it a try for giggles; if results are still no good then we can use a mutex. The performance hit should be pretty close to 0.

yakra:
### Heh.
https://github.com/TravelMapping/DataProcessing/blob/4879245fdfa727e57f90cd33b110a7ddd14e9bd6/siteupdate/cplusplus/classes/GraphGeneration/HighwayGraph.cpp#L512-L515
Turns out I was already using  the `+` instead of `<<` operator from the get-go.

yakra:
lab3's scaling above ~6 looks fishy. Changed too many variables at once? Test with...
* [ ] **InProg:** new timekeeping, no early termination
* [ ] **ToDo:** old timekeeping, early termination

yakra:
> ### Heh.
[...]

PR #311, opened 2020-04-17
concurrency augment speed & parallelism

yakra:
Closes https://github.com/yakra/DataProcessing/issues/117.

As we saw in #284, doing too much stuff while a mutex is locked can be a big hit to performance.
Similar to @jteresco's [suggestion](https://github.com/TravelMapping/DataProcessing/issues/281#issuecomment-592768745) in #281, we build `numthreads` lists of log entry lines, eliminate the mutex completely, and output them all single-threaded (taking ~0.2-0.3s) when done. A big win for parallelism.

Checking whether `hs != s` can speed us up a little in both languages.

**Python:**
* 9.2% faster on lab1
* 9.8% faster on lab1.5

**C++:**
ConcAugGood

### DIFF
--- a/siteupdate/cplusplus/siteupdate.cpp
+++ b/siteupdate/cplusplus/siteupdate.cpp
@@ -619,26 +619,30 @@ int main(int argc, char *argv[])
         //#include "debug/concurrency_augments.cpp"
       #ifdef threading_enabled
 	// set up for threaded concurrency augments
+	list<string>* augment_lists = new list<string>[args.numthreads];
 	list<TravelerList*>::iterator tl_it = traveler_lists.begin();
 
 	for (unsigned int t = 0; t < args.numthreads; t++)
-		thr[t] = new thread(ConcAugThread, t, &traveler_lists, &tl_it, &list_mtx, &log_mtx, &concurrencyfile);
+		thr[t] = new thread(ConcAugThread, t, &traveler_lists, &tl_it, &list_mtx, &log_mtx, augment_lists+t);
 	for (unsigned int t = 0; t < args.numthreads; t++)
 		thr[t]->join();
+	cout << "!\n" << et.et() << "Writing to concurrencies.log." << endl;
 	for (unsigned int t = 0; t < args.numthreads; t++)
+	{	for (std::string& entry : augment_lists[t]) concurrencyfile << entry << '\n';
 		delete thr[t];//*/
+	}
+	delete[] augment_lists;
       #else
 	for (TravelerList *t : traveler_lists)
 	{	cout << '.' << flush;
 		for (HighwaySegment *s : t->clinched_segments)
 		  if (s->concurrent)
 		    for (HighwaySegment *hs : *(s->concurrent))
-		      if (hs->route->system->active_or_preview() && hs->add_clinched_by(t))
+		      if (hs != s && hs->route->system->active_or_preview() && hs->add_clinched_by(t))
 			concurrencyfile << "Concurrency augment for traveler " << t->traveler_name << ": [" << hs->str() << "] based on [" << s->str() << "]\n";
 	}
+	cout << '!' << endl;
       #endif
-
-	cout << "!\n";
 	concurrencyfile.close();
 
 	/*ofstream sanetravfile(args.logfilepath+"/concurrent_travelers_sanity_check.log");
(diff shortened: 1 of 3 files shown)


yakra:
OK, this is pretty interesting...
This pull request did not change any of the code involved in `ComputeStatsThread`.
[ComputeStatsThread.cpp](https://github.com/TravelMapping/DataProcessing/commits/master/siteupdate/cplusplus/threads/ComputeStatsThread.cpp) is unchanged since Jun 23, 2019.
[compute_stats.cpp](https://github.com/TravelMapping/DataProcessing/blob/master/siteupdate/cplusplus/classes/HighwaySegment/compute_stats.cpp) is unchanged since May 19, 2019.
But as of 8a0b5fe9d7cc6664252dded823cb743fc5f91df1, **lab1** is running a tiny bit slower *Computing stats*, compared to the previous heads:

thr | ef1894f3f57c251089781862db7bd358aa1418dd | 166bc2c08340557c30d8bd92589fcd3ad4e370dd | e7b3c3edb6ff5967f43119525bbfc0bfd4376702 | cb5270a5e92db6014463fd18c3bd05edb8897c48 | 67bc8529d3aec35a7b876016beccce1916a045dd | a91cf27d48d1a75eae05dae7113de08aa6c8320c | 21f3ec9e18ffe5037764799b632c3fc3b7f53abc | 8a0b5fe9d7cc6664252dded823cb743fc5f91df1 | 8147a1fac0cd2d788c44c6a0255fdd0fe7e1e6c0 | 14ea929383adef4b52e2d3e4d1666b369ee180cc
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
**1** | 3.15 | 3.18 | 3.24 | 3.05 | 3.19 | 3.17 | 3.14 | **3.45** | 3.48 | 3.47
**2** | 1.71 | 1.76 | 1.75 | 1.73 | 1.7 | 1.68 | 1.69 | **1.86** | 1.91 | 1.91
**3** | 1.3 | 1.33 | 1.29 | 1.2 | 1.32 | 1.35 | 1.3 | **1.43** | 1.4 | 1.4
**4** | 1.31 | 1.29 | 1.28 | 1.13 | 1.35 | 1.31 | 1.29 | **1.45** | 1.39 | 1.38
**5** | 1.32 | 1.36 | 1.31 | 1.28 | 1.36 | 1.44 | 1.41 | **1.46** | 1.44 | 1.44
**6** | 1.35 | 1.33 | 1.34 | 1.28 | 1.38 | 1.39 | 1.39 | **1.45** | 1.45 | 1.5
**7** | 1.39 | 1.32 | 1.4 | 1.35 | 1.48 | 1.39 | 1.45 | **1.5** | 1.41 | 1.43
**8** | 1.3 | 1.34 | 1.35 | 1.31 | 1.37 | 1.39 | 1.44 | **1.5** | 1.43 | 1.43

Not *much* slower, mind you, just barely above the threshold to be consistent & noticeable.
**What's changed** is what happens *before* Computing stats: *Augmenting travelers for detected concurrent segments.* We're creating & storing a load of strings (~60 MB worth) that we weren't before.
**What I'm guessing** is that this pushes a load of other data out of L3 cache, to be re-accessed from main memory later on.
Seemingly confirming this, all I have to do is comment out [this line](https://github.com/TravelMapping/DataProcessing/blob/8a0b5fe9d7cc6664252dded823cb743fc5f91df1/siteupdate/cplusplus/threads/ConcAugThread.cpp#L20) and replace it with `int foo = 1`, and I get the lost speed back.
**What's the data in question?** In both `ConcAugThread` and `ComputeStatsThread`, we're iterating through series of `HighwaySegment` pointers. They're loaded from different places though; from `TravelerList::clinched_segments` and `Route::segment_list` respectively.
**What do we do *with* them?** In both cases, we iterate through their `concurrent` lists, and for all *those* segments, we access `route->system->level`, to check whether `route->system->active_or_preview()`.

...So it must be in there somewhere. I think. Maybe. Pretty crazy, right?
I'm running some tests, 21f3ec9e18ffe5037764799b632c3fc3b7f53abc vs. 8a0b5fe9d7cc6664252dded823cb743fc5f91df1, on lab2, which has 20MB L3 cache rather than 8, to see how that compares. The results won't be in while I'm awake.

PR #312, opened 2020-04-18
add repo head info to siteupdate.log

yakra:


jteresco:
Now that's Unix scripting, grabbing that info from git!

jteresco:
And successful on this afternoon's bonus site update.

PR #357, opened 2020-11-16
US_LETTER datacheck

### LINKED ISSUE #351: USA-only datachecks
* [x] **C++**: Rather than have each datacheck function check whether country == USA, check once then call all the functions.
* If I really wanna get jiggy wit it, no need to dereference a bunch of pointers and then iterate thru & compare up to 3 chars of the country string when we know beforehand whether a route's system is in the USA, and can set a `bool USA_flag`...
  - [x] **C++**: construct in Route::read_wpt; no need to recalculate for every point (easier, slightly less efficient)
  - [x] **C++**: pass to  Route::read_wpt from main & ReadWptThread; no need to recalculate for every route (more involved, more efficient)
  - [x] **Python:** ...in the outermost datacheck loop
* [x] Resurrect US_BANNER (as US_LETTER?)
  ```sh
  for rg in `ls | grep -v _`; do grep -sv '^+' $rg/*/*.wpt | cut -f1 -d' ' | egrep ':\*?US[0-9]+[AB][/_(]|:\*?US[0-9]+[AB]$'; done
  ```
  - [x] Python
  - [x] C++
  - [x] descriptor table
  - [x] datacheck_always_error
  - [x] wptedit
  - [x] syserr.php
  - [x] datacheck.php
  - [x] add "letter " [here](https://travelmapping.net/devel/manual/wayptlabels.php#bannerafternumber)
* [x] INTERSTATE_NO_HYPHEN does not account for leading `*`

yakra:
Closes #351.
See also https://github.com/TravelMapping/Web/pull/554.

The US_BANNER datacheck was present in siteupdate but commented out since the earliest days. Very few results overall (31 as of [3d2f858](github.com/TravelMapping/HighwayData/tree/3d2f858b7e86054cb3e6dc4aa8284fd2c8854dde)), though most were FPs, due to the A or B beginning a string such as "ALT" (vice "Alt") or a city suffix that would more properly go after an underscore.
This revision tightens the net, to omit these cases better left to a different datacheck.
Now, an error is flagged if the `A` or `B` ends the label or is followed by `/`, `_` or `(`.
Only **6 results.**

Renaming to US_LETTER: a better fit IMO, as most datacheck names describe what's wrong, whereas US_BANNER is more of a description of what would be right instead.

Along for the ride: some stuff. Commit messages & diffs should be self explanatory enough. :)

### DIFF
--- a/siteupdate/cplusplus/classes/DatacheckEntry.cpp
+++ b/siteupdate/cplusplus/classes/DatacheckEntry.cpp
@@ -3,9 +3,9 @@ class DatacheckEntry
 
     route is a pointer to the route with a datacheck error
 
-    labels is a list of labels that are related to the error (such
-    as the endpoints of a too-long segment or the three points that
-    form a sharp angle)
+    label1, label2 & label3 are labels that are related to the error
+    (such as the endpoints of a too-long segment or the three points
+    that form a sharp angle)
 
     code is the error code | info is additional
     string, one of:        | information, if used:
--- a/siteupdate/cplusplus/classes/DatacheckEntry.cpp
+++ b/siteupdate/cplusplus/classes/DatacheckEntry.cpp
@@ -36,7 +36,7 @@ class DatacheckEntry
     NONTERMINAL_UNDERSCORE |
     OUT_OF_BOUNDS          | coordinate pair
     SHARP_ANGLE            | angle in degrees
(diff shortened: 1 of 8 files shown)


jteresco:
Pulled to noreaster before tonight's update.

PR #367, opened 2020-12-02
show updated routes in user logs

### LINKED ISSUE #189: Produce a "things to check out" in user logs?
Thinking about some of the emails I get from users when they send in .list updates, where they find the errors in their lists from highway data updates over the course of a few days, it might be nice to have a report for users indicating which routes they have in their lists that have recent updates entries.

yakra:
closes #189

* short notices e.g. `Route updated 2020-12-01: CZE D6` at beginning of userlog, in line with errors msgs, limited to routes newly updated since the last .list update
* long list at end of userlog, listing dates & full update text for all .listed routes, sorted by date.

---

@jteresco, I also have a `siteupdate_centos.sh` for use in `python-teresco/`.
It's just what one would expect, the FreeBSD->CentOS changes without the Python->C++ changes.
Any objections if I commit & push, making it officially part of the repo?


michihdeu:
@yakra

- [ ] A notice on the forum to make users aware of it?
- [ ] Modify the manual? Maybe only a new issue first? https://travelmapping.net/participate.php?u=michih&#errorlog

jteresco:
> @jteresco, I also have a `siteupdate_centos.sh` for use in `python-teresco/`.
> It's just what one would expect, the FreeBSD->CentOS changes without the Python->C++ changes.
> Any objections if I commit & push, making it officially part of the repo?

No objections.  Maybe we can later put in some logic so one script works in both situations.

yakra:
Already putting this to use. I never unclinched the Tappan Zee Bridge.

yakra:
Let's not make a notice on the forum yet. There's a minor bug in localupdate.sh that causes the short list to show only "today's" updates.
https://github.com/TravelMapping/DataProcessing/blob/a905d2a5bdabb20ce5cfe25a28777f401c866d43/siteupdate/python-teresco/localupdate.sh#L71
**->**
```sh
 for u in *; do echo $u `git log -n 1 --pretty=%ci` $u; done > $execdir/listupdates.txt 
```

PR #368, opened 2020-12-04
route update fixes 

yakra:
* [route names from updates.csv rather than Route object](https://github.com/TravelMapping/DataProcessing/issues/189#issuecomment-738286167)
* [localupdate.sh bugfix: git log for each individual .list](https://github.com/TravelMapping/DataProcessing/pull/367#issuecomment-738336468)


jteresco:
This one's in and I'm running an extra site update now.

Issue #378, opened 2020-12-12
C++ code organization for separate compilation


jteresco:
Coming at this from my perspective as an experienced C programmer but someone whose last real C++ work was over 20 years ago, I am not sure this is an important issue.  However, the code organization with a single `.cpp` file that `#include`s everything else seems to be an unusual approach for a program of this size.

I am wondering if we should aim to reorganize some of the code to allow separate compilation and to eliminate `#include`s of `.cpp` files.  In C projects, I always organize code so that each `.c` file gets compiled to a `.o` file and only includes structure definitions, function prototypes, and `#define`s from any relevant `.h` files.  Then at the end, they're all linked to create the executable.

@yakra I'm not proposing any changes here yet, and certainly not asking you to make them, but would be interested in your thoughts as someone who's (obviously) done C++ coding decades more recently than I have.

yakra:
I may have hacked C++ as recently as last night, but that doesn't mean I'm doing things right or following all best practices.
I'm mainly self-taught, never really got the hang of linking or any related stuff, and not being forced to learn it, just kept on doing what worked for me in the beginning -- one simple painless g++ command.

* What are the advantages to this approach?
* Could there be performance penalties relative to the "one giant executable" approach used now?
* One thing I like to do is keep several executables around and have a shell script automate several hundred passes through them. This is how I get performance data and see how things scale with different numbers of threads, and make the graphs included in some issues & pull requests. This is easy with a self-contained executable for each revision of the code, but would it still be feasible if we're relying on separate object files?

jteresco:
You'd still have a single executable.  It's all in how it's compiled and linked.  However if you have different executables with different parameters compiled in, it would mean any object files that depend on those parameters would also need to be recompiled.

From a C programmer's perspective, one advantage to separate compilation of each C file comes in that you're guaranteed only a single version of each function and a single declaration of each global variable (if you have any) which might not happen if C files start to include each other.  In header files, you'd have only the function prototypes, which can be redefined as much as you want, and any global variable declaration would have `extern` in the header and only be declared "for real" in one C file.

Another thing that's nice, but not a big thing for a relatively small project like this in the era of fast compile times is that combined with a nice `Makefile`, any changes to code would only result in recompilation of those files that have changed (or which include `.h` files that have changed), and then a relink.  This used to be a big deal for large-scale projects where a 20-minute full recompile could be replaced with a few seconds to recompile a file or two that changed and relink.

[...]

PR #392, opened 2020-12-27
org for separate compilation + more

### LINKED ISSUE #327: WaypointQuadtree::points -- forward_list -> list
This is super obscure, and not too big a deal in the real world.
Almost a year and a half since #139 and this is the first I've noticed any issues.
On all but one of my machines, with older GCC versions, my code runs as *I* intended. On lab1.5, my code runs as *the C++ standard* intends.
[In April 2019, GCC was patched to make forward_list::sort stable](https://github.com/gcc-mirror/gcc/commit/a0295ab5313ec37e6f7963c9881a71e8c8a5f513).
`WaypointQuadtree` stores `points` in a `forward_list` container. This means, all this time, to get good DIFFs vs the Python version, I've been relying on a buggy compiler. I've been relying on my code working incorrectly. Should obviously fix that. :)

Using a stable sort with `forward_list` breaks things because its elements are essentially inserted backwards (`forward_list` has a `push_front` function, but no `push_back`).

* [x] So it looks like a regular `list` is what I want.
A couple functions will accordingly need adjustment...
  * [x] WaypointQuadtree::point_list
  * ~WaypointQuadtree::near_miss_waypoints~ not actually necessary.
  * ~(anything else?)~
* [x] Sorting may be slightly faster, as exit numbered waypoints and sequentially +Xnumbered shaping points will usually be sorted already. The effect is almost certain to not be noticeable, within margin of error.
  **Edit:** Noticeably slower (2.03s 496e840 vice 1.83s 362587e; HighwayData @ `b435cbd` & UserData @ `7704317`) `Sorting waypoints in Quadtree`. I'll attribute this to increased overhead in sorting a doubly linked list. No change `Sorting graph index waypoints by name priority`, which is already doubly linked.
* [x] OTOH, more overhead in constructing a `list` as compared to a `forward_list` could make *Reading waypoints for all routes* a bit slower. *Noticeably?* Certainly not *much*. **Edit:** No noticeable difference.
* Slightly larger RAM footprint too, but I think we can spare <8MB. ;)

---
How'd I notice; what was the symptom?
* `WaypointQuadtree::graph_points` processed [these 2 points](https://github.com/TravelMapping/HighwayData/blob/6d444b2c3534c8dc20f7cbc7b7ea2f219ebf17fa/hwy_data/AL/usaal/al.al154.wpt#L12-L13) in different orders, resulting in changing which vertex was named `AL154@+x06` and which was named `AL154@+x06|AL` in the region, country, continent, and master simple graphs -- waypointsimplification.log was not affected.

yakra:
* Commits thru eb1db56 on Dec. 22 work on organizing the Makefile and C++ code for separate compilation, per #378. It may not be our final destination, but it's a step ahead of what's in `master` right now.
* Close https://github.com/yakra/DataProcessing/issues/89. Both C++ and Python.
* Comment out WaypointQuadtree.tmg. It may have helped make sure I was on the right track when siteupdate.cpp was first crawling out of the muck, but has long outlived its usefulness. It may take 0.1s on Ubuntu or FreeBSD, but wastes 10-20s on CentOS depending on which machine.
* Close #327.


jteresco:
Meant to bring this in before tonight's update, but then realized I hadn't when the update was about 80% done.

yakra:
Nothing Earth-shattering on the Python side. We just save 0.7s *Sorting colocated point lists.*

PR #399, opened 2021-01-10
remove unused HGVertex member variables

yakra:
Remove sets populated during vertex construction but never used.
* Saves > 82 MB RAM in C++ version.
* A decent speedup `Creating unique names and vertices`:
  * \>23% C++ on lab2 (4.3s -> 3.48s)
  * \~44% Python on noreaster (17.6s -> 12.2s)


jteresco:
Trying to think of what those were supposed to be used for.  Maybe they were just added in anticipation of needing them for something?

yakra:
I assume so. I followed the blame back to when they were first introduced, but didn't take too close a look at the surrounding code to see if they were used.

yakra:
I've got some upcoming C++ changes (\~25% speedup writing subgraphs) that follow from this commit.
If this is something really worth saving, I could cherrypick or do an interactive rebase.
Or, if the need should ever arise in the future, it should be easy enough to add back in.

jteresco:
Sounds good to kill them off, then.  Merging.

PR #410, opened 2021-02-05
program to make deterministic/DIFFable TMGs

yakra:
Future plans include multithreaded processing of whole directories.


yakra:
Got too gung ho with new changes to this old program. There's a new bug where collapsed vertices get dropped.

PR #416, opened 2021-02-16
HgVertex cleanup/speedup; LABEL_INVALID_CHAR fixes

yakra:
* closes yakra#107
  * 3f938cbc0339c9ddf9bd101efd1c3dfa3a36ce9a cleans up vertex iteration. Dict keys are Waypoint objects rather than vertex labels now, and can be used for the HIDDEN_JUNCTION datacheck rather than the vestigial `first_waypoint` variable.
  * A [longstanding comment](https://github.com/TravelMapping/DataProcessing/blob/79bf1cda27cedb02d1d04408100af38fc2676743/siteupdate/python-teresco/siteupdate.py#L1692-L1693) notes that we can remove either `lat` & `lng` or `first_waypoint` from HGVertex, but not both. 315f834302e66636c5002678cba8262cb53ec2c0 removes `first_waypoint`. This option takes up about 4 MB more RAM, but the other option is slightly slower and requires minor changes to `PlaceRadius`.
* closes yakra#153
  * ac65bc389b28c48eaef53554ec494872d2165888 constructs & stores .tmg vertex line strings only once, retrieving them to write to all 3 graph files. This saves a surprising amount of time, subgraphs down to 67.23 s from 74.77 s on lab1. Python string construction gets pretty expensive what with interning, automatic hashing, etc.
* closes https://github.com/TravelMapping/HighwayData/issues/4561
  * b155944c33c1b5b4470bf1b1b16a42a54fff79ea flags LABEL_INVALID_CHAR only once per waypoint label, and notes cases caused by a UTF-8 BOM.
  * See also: https://github.com/TravelMapping/Web/pull/630

### DIFF
--- a/siteupdate/cplusplus/classes/GraphGeneration/HGVertex.cpp
+++ b/siteupdate/cplusplus/classes/GraphGeneration/HGVertex.cpp
@@ -19,9 +19,6 @@ HGVertex::HGVertex(Waypoint *wpt, const std::string *n, unsigned int numthreads)
 	    // 0: never visible outside of simple graphs
 	    // 1: visible only in traveled graph; hidden in collapsed graph
 	    // 2: visible in both traveled & collapsed graphs
-	// note: if saving the first waypoint, no longer need
-	// lat & lng and can replace with methods
-	first_waypoint = wpt;
 	if (!wpt->colocated)
 	{	if (!wpt->is_hidden) visibility = 2;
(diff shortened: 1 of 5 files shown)


yakra:
Might have one more commit to add on to this.

jteresco:
bring in?

yakra:
Ready to bring in. I'll save that commit for later.

Issue #433, opened 2021-06-08
VISIBLE_HIDDEN_COLOC & HIDDEN_JUNCTION w/only devel systems


yakra:
This just randomly occurred to me, not even looking at the code...
* VISIBLE_HIDDEN_COLOC check is performed in the `HGVertex` ctor.
* HIDDEN_JUNCTION check is performed during edge compression.

All of this is after the `HighwayGraph` ctor has been fed a list of every unique point that `is_or_colocated_with_active_or_preview()`, meaning these errors won't be flagged when only devel systems are involved.

* [x] (#435) Move VISIBLE_HIDDEN_COLOC check into `WaypointQuadtree::graph_points`
* [ ] HIDDEN_JUNCTION will require some more thought.

yakra:
> HIDDEN_JUNCTION will require some more thought.

Still not sure what exactly I wanna do here.
On the forum, @Duke87ofST [mentioned](https://forum.travelmapping.net/index.php?topic=5705.msg32170#msg32170) "the next point or the preceding point".
Something that takes this approach may be worth looking into.
It'll be a lot easier now that [TMArray is fully implemented](https://github.com/TravelMapping/DataProcessing/pull/607), `Waypoint`s & `HighwaySegment`s are stored sequentially, and pointer arithmetic can be used.

yakra:
To open flagged routes in the HB:
```sh
for line in `grep HIDDEN_JUNCTION logs/datacheck.log | fgrep -v -f <(grep HIDDEN_JUNCTION _f86f/logs/datacheck.log)`; do
  root=`echo $line | cut -f1 -d';'`
  label=`echo $line | cut -f2 -d';'`
  rteinfo=$(grep -h ";$root;" `ls ~/tm/HighwayData/hwy_data/_systems/*.csv | grep -v '_con\.csv$'`)
  sys=`echo $rteinfo | cut -f1 -d';'`
  rg=`echo $rteinfo | cut -f2 -d';'`
  coords=`grep "^$label" ~/tm/HighwayData/hwy_data/$rg/$sys/$root.wpt | sed 's~.*?~~'`
  echo "https://travelmapping.net/hb/showroute.php?r=$root&$coords&zoom=12"
done | xargs firefox
```

PR #435, opened 2021-06-09
VISIBLE_HIDDEN_COLOC for devel systems

yakra:
1st of 2 ToDo items for #433.
Refactors the VISIBLE_HIDDEN_COLOC datacheck to detect cases involving only devel systems.
In the process, cuts out some dead weight that's not been relevant a long time.

C++
Moved to a single-threaded part of the HighwayGraph ctor from a part that *can be* multi-threaded, but runs single-threaded by default. No big loss; `Setting up for graphs of highway data` takes 0.058s longer. :P


jteresco:
In for use in tonight's update.

PR #440, opened 2021-06-20
threaded stats CSVs / breaking the 30s barrier

yakra:
### AW YISS.
The 30-second barrier has been broken! A full `siteupdate -v` run, complete with nmp_merged files, subgraphs and DB file, took **29.6s** compiled with clang++ on lab1.5 (running Ubuntu).

### Threaded stats csv files
Most of the speedup benefits single-threaded operation -- Sometimes the *same damn code*, when refactored off into its own function, will magically run faster than when just included inline. Don't know why, but I've seen it happen. This is one of those times. With basically nothing changed, the stats csv file writing is >twice as fast. I'll take it!

There are 3 different components. Here they are with approximate times from BiggaTomato for illustrative purposes:
* **0.6s** allbyregionactiveonly.csv
* **0.9s** allbyregionactivepreview.csv
* **0.7s** all per-system csv files
* **2.2s** TOTAL

The approach here differs from other multi-threaded tasks.
With allbyregionactivepreview.csv taking up >1/3 of total time, there's no point in using >3 threads. So:
* **1** thread: Do allbyregionactiveonly.csv, allbyregionactivepreview.csv, all per-system csv files in order. Same as now.
* **2** threads:
  1. allbyregionactiveonly.csv, then share per-system csv files when done
  2. allbyregionactivepreview.csv, then share per-system csv files when done
* **3+** threads:
  1. allbyregionactiveonly.csv
  2. allbyregionactivepreview.csv
  3. all per-system csv files

Thus the benefits of the actual multi-threading are less pronounced.
* 0.3s dead time on one core + 0.2s on another while waiting 0.9s total for allbyregionactivepreview.csv to finish. Compared to a theoretical 0.73s for 1/3 the total time or 1.1s for 2 threads.
* In practice, things don't scale that well, even to 2 threads. 2 thr ~= 67-79% the total time of 1 thr, depending on what machine. Seems that there's a lot of data to pull from main memory wicked fast. Consistent with the systems with most RAM bandwidth (lab1, lab2) having the best scaling, and the systems with the least (lab3, BiggaTomato) actually performing worse with >1 thread. Accordingly, added a `-C` or `--st-csvs` commandline switch to disable multi-threaded stats csvs for systems with low memory bandwidth.

### DIFF
--- a/siteupdate/cplusplus/Makefile
+++ b/siteupdate/cplusplus/Makefile
@@ -1,10 +1,14 @@
 MTObjects = siteupdateMT.o functions/sql_fileMT.o threads/threads.o \
   classes/GraphGeneration/HighwayGraphMT.o \
-  classes/WaypointQuadtree/WaypointQuadtreeMT.o
+  classes/WaypointQuadtree/WaypointQuadtreeMT.o \
+  functions/allbyregionactiveonlyMT.o \
+  functions/allbyregionactivepreviewMT.o
 
 STObjects = siteupdateST.o functions/sql_fileST.o \
   classes/GraphGeneration/HighwayGraphST.o \
-  classes/WaypointQuadtree/WaypointQuadtreeST.o
+  classes/WaypointQuadtree/WaypointQuadtreeST.o \
+  functions/allbyregionactiveonlyST.o \
+  functions/allbyregionactivepreviewST.o
 
 CommonObjects = \
   classes/Args/Args.o \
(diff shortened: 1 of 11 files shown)


yakra:
Testing on noreaster...
* [ ] Observed some crashes & core dumps. This could be a reappearance of the other BSD bugs & crashes I've witnessed, or something specific to these changes. I'll do a post-mortem & see what I can see from the siteupdate.log files.
* [ ] I suspected that with noreaster's memory bus doing a lot of stuff, this would run slower multi-threaded, like on BiggaTomato or lab3. Confirmed. I'll make single-threaded the default behavior & multi-threaded an option.

PR #448, opened 2021-07-07
Operation Valgrind Parts 1&2

yakra:
Datacheck: delete fp before invalidating iterator
HGEdge: prevent memory leak from "reversed" edges
WaypointQuadtree: proper recursive destructor
main: prevent improper line.back() on empty nmpfps

---
TravelerList: don't "strip whitespace" before file BOF
ElapsedTime: plug memory leak on internal str buffer


yakra:
Comments need to be fixed, not just the code they're commenting on.
Going to rebase & force push. By then, more commits may be along for the ride too.

PR #451, opened 2021-07-19
concurrency detection / graphs

### LINKED ISSUE #248: 0-byte WPTs
0-byte WPT files crash the C++ siteupdate program.
I haven't tested siteupdate.py to see how it copes.

### LINKED ISSUE #449: concurrency detection memory leak
Valgrind reports:
* definitely lost: 24 B (clang) / 16 B (g++) in 1 block, allocated [here](https://github.com/TravelMapping/DataProcessing/blob/e7df093957fb974e23392fc83897be5bcfeb60a5/siteupdate/cplusplus/tasks/concurrency_detection.cpp#L19).
  An orphaned concurrency list is not attached to any HighwaySegment. It's created, partially populated, and then its pointer is overwritten during concurrency detection.
* indirectly lost: 144 B in 6 blocks (i.e., 6 `std::_List_node`s)
  * 24 B in 1 block, allocated [here](https://github.com/TravelMapping/DataProcessing/blob/e7df093957fb974e23392fc83897be5bcfeb60a5/siteupdate/cplusplus/tasks/concurrency_detection.cpp#L22), the 1st route when the concurrency is first detected.
  * 24 B in 1 block, allocated [here](https://github.com/TravelMapping/DataProcessing/blob/e7df093957fb974e23392fc83897be5bcfeb60a5/siteupdate/cplusplus/tasks/concurrency_detection.cpp#L23), the 2nd route when the concurrency is first detected.
  * 96 B in 4 blocks, allocated [here](https://github.com/TravelMapping/DataProcessing/blob/e7df093957fb974e23392fc83897be5bcfeb60a5/siteupdate/cplusplus/tasks/concurrency_detection.cpp#L28), when the concurrency is augmented with more routes afterward.
* My first guess was our old pal outside the Fort Pitt Tunnel, with 6 concurrent routes (I-376, US19, US22, US30, SB US19Trk, NB US19Trk).
* But nope! The missing object was first assigned to [POL E67 29(S8) 27(A1)](https://travelmapping.net/hb/showroute.php?r=pol.e67&lat=51.4230287273842&lon=19.638833999633793&zoom=13). Same deal here with a wrong-way self-multiplex -- the ramps for the desired E67/S8 movements aren't built yet, so traffic continues straight on through, U-turns at the next interchange, and doubles back, leaving via the other road. There are 7 concurrent segments here, processed in system>route order:
  1. `POL E67 29(S8) 27(A1)` The first list constructed gets built out to 6 of 7 routes, then its pointer is overwritten.
  2. `POL E67 27(A1) 26(A1)`
  3. `POL E75 26(A1) 27(A1)`
  4. `POL A1 26 27`
  5. `POL S8 26(A1) 27(A1)`
  6. `POL S8 27(A1) 29`
  7. `POL DK74 A1_S A1_N`
* Why does POL have a memory leak but not PA? Could be any of a number of different factors at play here. This will take a long walk thru the code like in #135...

### LINKED ISSUE #450: Simplifying collapsed graph file names
METAL treats the "collapsed" format graphs as the default in nearly all situations at this point.  There is no need for the "-collapsed" to be part of the filenames.  This should not require any significant changes (possibly none) in HDX, but some in the documentation.  The only change needed in the site update code should be to remove the extra text from collapsed-format filenames.

yakra:
closes #450
closes yakra#177
closes #449
closes #248

### DIFF
--- a/siteupdate/cplusplus/classes/Datacheck/Datacheck.cpp
+++ b/siteupdate/cplusplus/classes/Datacheck/Datacheck.cpp
@@ -88,8 +88,8 @@ void Datacheck::mark_fps(std::string& path, ElapsedTime &et)
 		    {	//std::cout << "Match!" << std::endl;
 			d.fp = 1;
 			fpcount++;
-			fps.erase(fp);
 			delete[] *fp;
+			fps.erase(fp);
 			break;
 		    }
 		    else
--- a/siteupdate/cplusplus/classes/Datacheck/Datacheck.cpp
+++ b/siteupdate/cplusplus/classes/Datacheck/Datacheck.cpp
@@ -56,7 +56,7 @@ void Datacheck::read_fps(std::string& path, ErrorList &el)
 		// parse datacheckfps.csv line
 		size_t NumFields = 6;
 		std::string* fields = new std::string[6];
-				 // deleted when FP is matched or on termination of program
+				 // deleted when FP is matched or when writing unmatchedfps.log
 		std::string* ptr_array[6] = {&fields[0], &fields[1], &fields[2], &fields[3], &fields[4], &fields[5]};
 		split(line, ptr_array, NumFields, ';');
 		if (NumFields != 6)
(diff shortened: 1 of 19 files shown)


yakra:
Adding 7 more commits that help make siteupdate more "Valgrind clean", avoid invalid reads & writes, and plug lower-impact memory leaks.

Class destruction for bigger-ticket items is still subject to change, so I'm holding off on pushing to master for now.

yakra:
## Python speed on lab1

task | before | after
| - | - | - |
Concurrent segment detection | 4.78 s | 2.98 s
Creating edges | 7.51 s | 7.08 s

Issue #377, opened 2020-12-12
C++ build process


jteresco:
I'd like to add a `Makefile` that can build the C++ site update program with appropriate compiler options based on the machine where it's running.

yakra:
`Make` is another thing I never learned my way around. I've always perceived it as something unimaginably complex & arcane, and have vague memories of trying to compile LAME when I was new to Linux, and eventually throwing up my hands.
OTOH, last year I followed the directions on the internet and compiled python3 no problemo.

What kind of options would we be looking at?
* Something to manage or take the place of the current `#define` to enable/disable threading?
* Optimizations for the machine where it's running, e.g. SIMD support, etc? I've always assumed g++ by default will optimize what it can, and not looked too closely at its options.
* Or are we talking more basic stuff like `-pthread` `-std=c++11`?

jteresco:
The `Makefile` basically just defines a bunch of rules for how to take some set of input files (source code), possibly through the creation/update of some intermediate files (object files), to get your output files (executables).

At its simplest, it would just define a single rule to run the one `g++` command.  Better, it would compile only what is needed in a separate compilation setup.  And, we could set it up to detect the system we're building on and run the right C++ compiler with the right flags.  And yes, I would have it so it either would always unconditionally build both a single- and multi-threaded version of the executable, or have a `make` command line parameter to build the one we wanted (something like `make THREADS=1`).

One technique I like is to have a `make` command-line parameter choose the appropriate `-D` flags to the compiler to essentially `#define` one or more things from the compiler line rather than having to edit code to change values.

In any case, I can add a simple one to get started that would really only mean instead of typing the whole `g++` line to compile, you would just type `make` (or on FreeBSD, probably `gmake` to run GNU Make).

yakra:
>  Better, it would compile only what is needed in a separate compilation setup.

Sounds like what you mentioned in https://github.com/TravelMapping/DataProcessing/issues/378#issuecomment-743782361 (more to come there shortly): *"any changes to code would only result in recompilation of those files that have changed (or which include .h files that have changed)"*.

> I would have it so it either would always unconditionally build both a single- and multi-threaded version of the executable, or

Sounds like a Good Thing. Convenient, too. Sometimes I'm hacking away at the threaded version and forget to even compile the single-threaded version and make sure nothing is broken...

[...]

PR #459, opened 2021-09-19
WaypointQuadtree::final_report

yakra:
([Anakin & Padme meme, large image](https://www.reddit.com/r/ProgrammerHumor/comments/pq0ia7/oh_the_leaks/))
Part of [making siteupdate "Valgrind clean"](https://github.com/TravelMapping/DataProcessing/issues/375#issuecomment-874435287) is explicitly [`delete`ing dynamically allocated memory. Compared to just letting the OS magically delete everything allocated](https://github.com/TravelMapping/DataProcessing/issues/375#issuecomment-875238626) when the program stops running, it's a bit expensive though.

I opted to keep costs down by leveraging some recursion, iteration & branching that's already happening anyway, and folding deletion of quadtree nodes, waypoints, and colcation lists into a new function that handles gathering/printing data for the final colocation stats report.
Along the way, getting rid of some redundant iteration, counting, list construction, etc. speeds things up, eliminating `WaypointQuadtree::max_colocated()` and `Waypoint::num_colocated()` in the process, changes which also make it into the Python version.
* **C++:** The colocation report itself is down from 0.91 to 0.2 s on BiggaTomato. Add in 0.825 s for the deletions, and we're only 0.115 s above where we started. 
* **Python:** Colocation reports are faster. No deletions added to the process though; that's still all handled by the garbage collector when the program terminates. The "during" version listed in the table below has all the changes except for `WaypointQuadtree.final_report` being carved out into its own function.

  Box | Before | During | After
  | - | - | - | - |
  BT | 2.5 | 1.2 | .8
  lab1 | 1.5 | .8 | .5
  lab1.5 | 1.3 | .7 | .5
  lab2 | 1.6 | .9 | .7
  lab3 | 2.1 | 1.2 | 1.0
  lab4 | 1.9 | 1.2 | .7

### DIFF
--- a/siteupdate/cplusplus/classes/GraphGeneration/HighwayGraph.cpp
+++ b/siteupdate/cplusplus/classes/GraphGeneration/HighwayGraph.cpp
@@ -59,7 +59,7 @@ HighwayGraph::HighwayGraph(WaypointQuadtree &all_waypoints, ElapsedTime &et)
 		  for (HighwaySegment *s : r->segment_list)
 		    if (!s->concurrent || s == s->concurrent->front())
 		      new HGEdge(s, this);
-		      // deleted by ~HGVertex, called by HighwayGraph::clear
+		      // deleted by HGEdge::detach via ~HGVertex via HighwayGraph::clear
 	}
 	std::cout << '!' << std::endl;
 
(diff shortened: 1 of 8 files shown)


jteresco:
This and #458 are on noreaster for production site updates.  Thanks.

michihdeu:
> This and #458 are on noreaster for production site updates. Thanks.

I was confused reading "production site update" in combination to @yakra's c++ work but I realized that @yakra also modified the python code. 😌 

Will we soon switch to c++ or will we stick with using the python script for the time being?

Just as an reminder, this will have to be updated once we will switch: https://github.com/TravelMapping/DataProcessing/blob/master/SETUP.md

yakra:
@michihdeu [wrote](https://github.com/TravelMapping/DataProcessing/pull/459#issuecomment-924251583):
> Will we soon switch to c++ or will we stick with the python script for the time being?

Soon, hopefully! The main thing holding us back from running the multithreaded C++ siteupdate program on noreaster is the need to [switch compilers from g++ to clang++](https://github.com/yakra/DataProcessing/issues/188). All that will need is some minor changes to the [Makefile](https://github.com/TravelMapping/DataProcessing/issues/377). I've had a new version kicking around locally for quite some time, but held off on committing it 'cause I wanted to do all my silly little benchmarks. :grin:
When I first looked into it, I was getting some results inconsistent from earlier benchmarking runs, so I put it off a while and got sidetracked by other things. Finally got the inconsistencies sorted out, and spent today *(uhh... yesterday)* arranging all the data & finishing off most of the charts (10 down, 2 or 3 to go). I'll finish those off then get the pull request in after sleeping.

Then, [a few more odds & ends before moving to production](https://github.com/TravelMapping/DataProcessing/issues/461).

After switching to clang, I want to spend some more time plugging the last of the memory leaks. This needn't *necessarily* hold us back from the C++ version moving to production, as all the memory it allocates is freed upon termination of the program, even on FreeBSD it would appear, but it's just good practice.

PR #465, opened 2021-10-02
g++ -> clang++

### LINKED ISSUE #377: C++ build process

yakra:
* Closes #377. Replacing gcc with clang is necessary to run multithreaded on FreeBSD.
  On Linux, The clang build performs comparable or better. I'm fine with keeping it simple, not worrying about conditional checks, and compiling with clang all the time.
  * `siteupdateST`: Some tasks are faster with clang, some slower.
    Overall, total run time is faster with clang, on all machines. By ~1s on my CentOS boxes, up to 4-6s on Ubuntu.
  * Multi-threaded `siteupdate` is a bit less clear-cut...:
    * The tasks that stay single-threaded in `siteupdate` appear to account for a smaller part of the time savings in `siteupdateST`. Seems pretty close to margin of error. Faster on some machines, slower on others.
    * Combined times for multi-threaded tasks are faster across the board on Ubuntu. On CentOS, higher thread counts lean toward gcc, and lower thread counts lean toward clang. Clock speed may play a role -- lab1 was faster in 6/8 trials, by 0.83 s average across all thread counts; lab2 was slower by 0.07 s, and lab3 by 0.24 s.
    * Total run time, ST & MT tasks together: Consistently faster on Ubuntu, by about 2.1 s. The only trial that was slower was BiggaTomato @ 4 threads, by 0.35 s. If I test it again maybe it'll be faster. :) On CentOS, lower thread counts favor clang; overall gcc is faster by .08 - .24 s ballpark, so a pretty slim margin.
* Closes yakra#188. That issue has enough speed measurements, tables and charts to make your eyes bleed.

### DIFF
--- a/siteupdate/cplusplus/Makefile
+++ b/siteupdate/cplusplus/Makefile
@@ -1,3 +1,6 @@
+CXX = clang++
+CXXFLAGS = -std=c++11 -Wno-comment -Wno-dangling-else -Wno-logical-op-parentheses
+
 MTObjects = siteupdateMT.o functions/sql_fileMT.o threads/threads.o \
   classes/GraphGeneration/HighwayGraphMT.o \
   classes/WaypointQuadtree/WaypointQuadtreeMT.o \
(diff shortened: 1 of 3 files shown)


yakra:
:+1:

PR #478, opened 2021-11-06
ABBREV_AS_CON_BANNER info for both csv line numbers

yakra:
See also: https://github.com/TravelMapping/Web/pull/698


jteresco:
On noreaster for tonight's update.

PR #479, opened 2021-11-06
MISSING_CITY and EXTRANEOUS_ABBREV

### LINKED ISSUE #303: ABBREV_NO_CITY
http://forum.travelmapping.net/index.php?topic=2138.msg18215#msg18215
> On second thought...
It's possible there could be some rare (but legitimate?) reasons we'd want to do this. Take [KS171Pit](http://travelmapping.net/hb/?r=ks.ks171pit). There used to be an ordinary KS171, but KDOT decommissioned it. It makes sense to remove the City field, as there's no other KS171 to differentiate from.
Could I & should I rename it KS171 & make KS171Pit an AltLabel in this case. Yeah. But the point is...
Imagine that the original plain KS171 became an AltRouteName for another route, and that it remained [in-use](https://github.com/TravelMapping/DataProcessing/issues/230) in a .list file. KS171Pit would by necessity have to remain KS171Pit.
>
> It makes sense to have these not be a showstopper that terminates the siteupdate program. They could go into datacheck, with the route flagged as per normal, just with no waypoint labels listed, maybe with the system name .csv in the "Info" column to provide a little clarity.
If we really wanna get fancy with it, we could check for the corresponding abbrev-less route in [in-use AltRouteNames](https://github.com/TravelMapping/DataProcessing/issues/230), and if found, not flag the error. I think this could even be a `datacheck_always_error`.

yakra:
Closes #303.
https://forum.travelmapping.net/index.php?topic=2138.msg18207#msg18207

### DIFF
--- a/siteupdate/cplusplus/classes/Datacheck/Datacheck.cpp
+++ b/siteupdate/cplusplus/classes/Datacheck/Datacheck.cpp
@@ -137,6 +137,7 @@ std::list<std::string*> Datacheck::fps;
 std::unordered_set<std::string> Datacheck::always_error
 {	"ABBREV_AS_CHOP_BANNER",
 	"ABBREV_AS_CON_BANNER",
+	"ABBREV_MISSING_CITY",
 	"BAD_ANGLE",
 	"CON_BANNER_MISMATCH",
 	"CON_ROUTE_MISMATCH",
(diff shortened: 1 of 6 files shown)


yakra:
Should maybe think about this one a little more.

yakra:
Reopened only so the new commit would show up.
Re-closing.

Issue #490, opened 2022-02-14
shell script: users affected by a relocation


yakra:
[**PUT FORUM LINK HERE**](https://forum.travelmapping.net/index.php?topic=4804.msg26915#msg26915)

**idea:** a shall script that, given a route section (input in standard 4-field .list line style format) that's been relocated in the middle of a route, finds all .list files affected. Even those using two "good" unchanged points on either side of the relocation.

* [ ] For both point labels in .list lines referencing this route, look up line numbers,
  ```sh
  file=HighwayData/hwy_data/ME/usame/me.me077.wpt #for example
  LabelTable=`cat -n $file | sed -r -e 's~^ *~~' -e 's~http.*~~' | tr -d '+*' | tr '\t' ' '`
  echo "$LabelTable" | grep " $ALabelFromSomeonesList " | cut -f1 -d' '
  ```
  find out whether before, within, or after relocated section
  * 1st point within -> flag
  * 2nd point within -> flag
  * before & before -> don't
  * before & after -> flag
  * after & before -> flag
  * after & after -> don't
* [ ] Invalid labels will yield an empty string for a line number. Skip over stuff.
* [ ] [6-field lines on one chopped route](https://github.com/TravelMapping/UserData/blob/484ce43b2eed0fe53f265917dd2c12a824d306f3/list_files/master_son.list#L364-L365) will be easy. Convert to 4-field using sed.
* [ ] Multi-region 6-field lines will be more complicated.

michihdeu:
I think that the algorithm might be used for it - if we want.

yakra:
Wait. **You're not wrong.**
At least, in the oversimplified way I was conceptualizing it.
It's still not something I want to pursue though.

**What I didn't consider:**
* Sure, this would produce false positives, but what we already have produces *even more* FPs.
* This would get rid of some (but not all) of the FPs we already have.

**However...**
* There'd need to be some way of distinguishing the most recent newsworthy update from other more recent non-newsworthy updates. This can be done, but it'd be a pain.
* Even if implemented successfully, it would be painfully slow. The info would have to be gotten via git, before launching siteupdate, the same way listupdates.txt (used for the `Route updated 2022-02-13: PA US13` log file entries) is created now.
Extrapolating from that, this could take about 6 minutes. Almost as much time as a full Py site update, even more time than a full C++ one.
And that figure is unrealistically optimistic because this process is more complicated:
[...]

PR #491, opened 2022-02-16
FLOAT -> DOUBLE in four SQL tables / flippedroutes.log

yakra:
**FLOAT -> DOUBLE in four SQL tables**
https://forum.travelmapping.net/index.php?topic=2065.msg26937#msg26937
* overallMileageByRegion
* systemMileageByRegion
* clinchedOverallMileageByRegion
* clinchedSystemMileageByRegion

---

Along for the ride:
flippedroutes.log lists routes that are flagged as reversed for processing ConnectedRoutes in 6-field .list entries.
It's very inexpensive to write, super-easy to code, and should come in handy when I want a quick reference for test cases to use in development.

### DIFF
--- a/siteupdate/cplusplus/functions/sql_file.cpp
+++ b/siteupdate/cplusplus/functions/sql_file.cpp
@@ -262,7 +262,7 @@ void sqlfile1
 	std::cout << et->et() << "...overallMileageByRegion" << std::endl;
       #endif
 	sqlfile << "CREATE TABLE overallMileageByRegion (region VARCHAR(" << DBFieldLength::regionCode
-		<< "), activeMileage FLOAT, activePreviewMileage FLOAT);\n";
+		<< "), activeMileage DOUBLE, activePreviewMileage DOUBLE);\n";
 	sqlfile << "INSERT INTO overallMileageByRegion VALUES\n";
 	first = 1;
 	for (Region* region : Region::allregions)
@@ -282,7 +282,7 @@ void sqlfile1
       #endif
 	sqlfile << "CREATE TABLE systemMileageByRegion (systemName VARCHAR(" << DBFieldLength::systemName
 		<< "), region VARCHAR(" << DBFieldLength::regionCode
-		<< "), mileage FLOAT, FOREIGN KEY (systemName) REFERENCES systems(systemName));\n";
+		<< "), mileage DOUBLE, FOREIGN KEY (systemName) REFERENCES systems(systemName));\n";
 	sqlfile << "INSERT INTO systemMileageByRegion VALUES\n";
 	first = 1;
 	for (HighwaySystem *h : HighwaySystem::syslist)
@@ -303,7 +303,7 @@ void sqlfile1
       #endif
 	sqlfile << "CREATE TABLE clinchedOverallMileageByRegion (region VARCHAR(" << DBFieldLength::regionCode
 		<< "), traveler VARCHAR(" << DBFieldLength::traveler
-		<< "), activeMileage FLOAT, activePreviewMileage FLOAT);\n";
+		<< "), activeMileage DOUBLE, activePreviewMileage DOUBLE);\n";
 	sqlfile << "INSERT INTO clinchedOverallMileageByRegion VALUES\n";
 	first = 1;
 	for (TravelerList *t : TravelerList::allusers)
@@ -327,7 +327,7 @@ void sqlfile1
 	sqlfile << "CREATE TABLE clinchedSystemMileageByRegion (systemName VARCHAR(" << DBFieldLength::systemName
 		<< "), region VARCHAR(" << DBFieldLength::regionCode
 		<< "), traveler VARCHAR(" << DBFieldLength::traveler
-		<< "), mileage FLOAT, FOREIGN KEY (systemName) REFERENCES systems(systemName));\n";
+		<< "), mileage DOUBLE, FOREIGN KEY (systemName) REFERENCES systems(systemName));\n";
 	sqlfile << "INSERT INTO clinchedSystemMileageByRegion VALUES\n";
 	first = 1;
 	for (std::string &line : clin_db_val->csmbr_values)
--- a/siteupdate/cplusplus/siteupdate.cpp
+++ b/siteupdate/cplusplus/siteupdate.cpp
@@ -401,13 +401,14 @@ int main(int argc, char *argv[])
 		r->duplicate_labels.clear();
 	  }
 
-	cout << et.et() << "Writing pointsinuse.log, unusedaltlabels.log, listnamesinuse.log and unusedaltroutenames.log" << endl;
+	cout << et.et() << "Writing route and label logs." << endl;
 	unsigned int total_unused_alt_labels = 0;
 	unsigned int total_unusedaltroutenames = 0;
 	list<string> unused_alt_labels;
 	ofstream piufile(Args::logfilepath+"/pointsinuse.log");
 	ofstream lniufile(Args::logfilepath+"/listnamesinuse.log");
 	ofstream uarnfile(Args::logfilepath+"/unusedaltroutenames.log");
+	ofstream flipfile(Args::logfilepath+"/flippedroutes.log");
 	timestamp = time(0);
 	piufile << "Log file created at: " << ctime(&timestamp);
 	lniufile << "Log file created at: " << ctime(&timestamp);
@@ -433,6 +434,8 @@ int main(int argc, char *argv[])
 				unused_alt_labels.push_back(ual_entry);
 				r->unused_alt_labels.clear();
 			}
+			// flippedroutes.log line
+			if (r->is_reversed) flipfile << r->root << '\n';
 		}
 		// listnamesinuse.log line
 		if (h->listnamesinuse.size())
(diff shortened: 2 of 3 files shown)


yakra:
Good to go.

jteresco:
And now a site update using this has completed.

yakra:
I looked at https://travelmapping.net/logs/flippedroutes.log just for giggles. Chromium thinks it's in Scottish Gaelic. Attempting to translate to English changes nothing. Giggles confirmed.

michihdeu:
> I looked at https://travelmapping.net/logs/flippedroutes.log

I never realized flippedroutes.log. What does it mean? It lists routes I maintain and I'd like to know if anything is wrong with them.... 😄

yakra:
It's brand new with this pull request.
From the OP:
> Along for the ride:
flippedroutes.log lists routes that are flagged as reversed for processing ConnectedRoutes in 6-field .list entries.
It's very inexpensive to write, super-easy to code, and should come in handy when I want a quick reference for test cases to use in development.

There won't be anything wrong with these routes unless flagged [DISCONNECTED_ROUTE](https://travelmapping.net/devel/datacheck.php?show=DISCONNECTED_ROUTE).

I guess I should put a blurb under [Web Site Development and Site Update Logs](https://travelmapping.net/devel/tools.php)

michihdeu:
> There won't be anything wrong with these routes unless flagged [DISCONNECTED_ROUTE](https://travelmapping.net/devel/datacheck.php?show=DISCONNECTED_ROUTE).

but why are they listed?

yakra:
> > should come in handy when I want a quick reference for test cases to use in development.

Or, it can help streamline shell scripts.
For example, for the one in #490 to be fully functional it has to accommodate flipped routes for multi-region .list entries.
I could have the shell script re-create all the [same](https://github.com/TravelMapping/DataProcessing/blob/38da77b/siteupdate/cplusplus/classes/HighwaySystem/route_integrity.cpp#L55-L78) [steps](https://github.com/TravelMapping/DataProcessing/blob/4fd84a4/siteupdate/python-teresco/siteupdate.py#L3028-L3045) siteupdate does to flag a route as reversed -- a moderate PitA from scratch in bash -- or I could just say
```sh
if [ `curl https://travelmapping.net/logs/flippedroutes.log | grep -c ^$root$` == 1 ];
  then thing=$stuff
  else thing=$SomethingElse
fi
```
Much simpler!

> but why are they listed?

Or do you mean, why are they flagged as reversed in the first place?
For some routes, the point order in the chopped route mismatches the root order in the connected route. For example,
`MO US275` [chopped](https://travelmapping.net/hb/showroute.php?r=mo.us275) | [connected](https://travelmapping.net/hb/showroute.php?r=mo.us275&cr)
`PA I-295` [chopped](https://travelmapping.net/hb/showroute.php?r=pa.i295) | [connected](https://travelmapping.net/hb/showroute.php?r=pa.i295&cr) (note how for the connected route, the PA exits count down from 10 to 3)
https://forum.travelmapping.net/index.php?topic=3648.msg19036#msg19036
https://forum.travelmapping.net/index.php?topic=3652.msg19105#msg19105

michihdeu:
> There won't be anything wrong with these routes unless flagged [DISCONNECTED_ROUTE](https://travelmapping.net/devel/datacheck.php?show=DISCONNECTED_ROUTE).

No, they are listed because there IS something wrong although they are NOT flagged DISCONNECTED_ROUTE!
I fixed the routes mentioned in flippedroutes.log but:

Reported correctly:
```
frages.n019
frabfc.n019
frahdf.d100160
fraidf.d100160
frahdf.d100160cha
```

Reported incomplete:
`frages.d0002f88mar`
because the other chopped routes are **not listed - WHY?**:
```
frages.d0002f88
frabfc.d0002f88
```

Cool data check though. Love it 😄

yakra:
I think I see what you mean here -- there's something wrong with the routes stylistically, from a consistency perspective; their points (or connected roots) are just plain in the wrong order for that region and/or system and should be reversed in the HighwayData itself. *(Right?)* Fair enough.

My threshold for "something wrong" was a datacheck error that'd potentially affect how users' travels are calculated from a .list line.
* If there's no DISCONNECTED_ROUTE error, the software flags a route as "reversed" behind the scenes, preventing stuff like [this](https://forum.travelmapping.net/index.php?topic=3648.msg19036#msg19036), and life goes on.
* If there is a DISCONNECTED_ROUTE error, [all bets are off; a multi-region .list entry may or may not work correctly](https://github.com/TravelMapping/DataProcessing/pull/495).

> Cool data check though. Love it 😄

My intent wasn't for this to be a datacheck per se (though you can certainly use it as one if you wish), more of just a handy-dandy quick reference to use for development purposes (thus I think it should be listed [here](https://travelmapping.net/devel/logs.php#weblogs) rather than [here](https://travelmapping.net/devel/logs.php#nmplogs)).

So, for a route to be reported "correctly" in flippedroutes.log, it just has to be one of the routes flagged behind the scenes as reversed, for having its point order opposite the connected root order.

> ```
> frages.n019
> frabfc.n019
> frahdf.d100160
> fraidf.d100160
> frahdf.d100160cha
> ```

N19 (Vesoul) had its chopped roots ordered east-to-west, with the point order therein west-to-east.
D1001-60 had its chopped roots ordered north-to-south, with the point order therein south-to-north.
I noticed that you [changed the _con.csv order](https://github.com/TravelMapping/HighwayData/pull/5618/commits/3ac3541a54281ed0b5b4e6bb17c1dc432a06012c) for these. Another solution that would have "fixed" these cases would have been to reverse the point order in the .wpt files instead. You know what's appropriate for France; I don't. (What's normal for France? S->N or N->S? W->E or E->W?)

> Reported incomplete:
> `frages.d0002f88mar`
> because the other chopped routes are **not listed - WHY?**:
> ```
> frages.d0002f88
> frabfc.d0002f88
> ```

The chopped roots were listed east-to-west.
* `frages.d0002f88` & `frabfc.d0002f88` have their points ordered east-to-west, same as the connected root order. They don't get flagged as reversed, and don't get put in the log file.
* Only `frages.d0002f88mar` has its points west-to-east, so it's the only only chopped route flagged as reversed.

> There are more flipped reports where only one chopped route is listed.

Yes, I expect that this will be normal behavior, what we'll see *most* of the time, especially if there are no "errors" (i.e., routes ***un***intentionally flipped) present.

michihdeu:
I always order routes N-S and W-E but France is usually Paris centric (since N routes were drafted by Si)

I initially dealed wrong with GES 2F but think that I got it now 😄 Thanks!

michihdeu:
> My intent wasn't for this to be a datacheck per se (though you can certainly use it as one if you wish), more of just a handy-dandy quick reference to use for development purposes (thus I think it should be listed [here](https://travelmapping.net/devel/logs.php#weblogs) rather than [here](https://travelmapping.net/devel/logs.php#nmplogs)).

It is not a nmplog but a general log. yep, go for it! I'd say between unprocessedwpts.log and https://travelmapping.net/graphs/ I could also do it but don't know how to describe it properly because of the two causes for routes being listed.

yakra:
> I always order routes N-S and W-E but France is usually Paris centric (since N routes were drafted by Si)
> I initially dealed wrong with GES 2F but think that I got it now 😄 Thanks!

I think D1001 should have also hade the points reversed rather than the routes, to go N-S.

> > My intent wasn't for this to be a datacheck per se (though you can certainly use it as one if you wish), more of just a handy-dandy quick reference to use for development purposes (thus I think it should be listed [here](https://travelmapping.net/devel/logs.php#weblogs) rather than [here](https://travelmapping.net/devel/logs.php#nmplogs)).

Oops, I wrote that post too fast, based on having looked at [*tools.php*]https://travelmapping.net/devel/tools.php) before. I meant to say "[here](https://travelmapping.net/devel/logs.php#weblogs) rather than [here](https://travelmapping.net/devel/logs.php#hwylogs)"

> It is not a nmplog but a general log. yep, go for it! I'd say between unprocessedwpts.log and https://travelmapping.net/graphs/ I could also do it but don't know how to describe it properly because of listed.

So yes, definitely not an NMP log.
While my intent for this was as more of a **Web Site Development and Site Update Log**, I could see the logic in listing it as a **Highway Data Development Log** instead. You've certainly made use of it for such purposes.
There's some stuff in caused by data errors, and some stuff in it that's false positive, as is often the case.

> the two causes for routes being

Sounds like you're thinking of...?
* reversed roots in _con.csv file
* reversed points in .wpt file

This can also be conceptualized as just one cause:
* the order of the roots in the _con.csv file and the order of the points in the .wpt file don't match.

*Shrug* -- Anyway!

> EDIT: Maybe...
>
> ```
> List of flipped .wpt files
> Waypoint files that are flagged as reversed for processing ConnectedRoutes in 6-field .list entries.
> ```

Yeah, you get the idea, this is good.
If listing this as a Web Site Development and Site Update Log, maybe just use the bottom line, to keep with single-line descriptions.
If listing this as a Highway Data Development Log, use both lines, and keep the same visual presentation.

Though if I wanna get pedantic, I might say "multi-region" instead of "6-field" .list entries -- [6-field entries listing a single chopped route](https://github.com/TravelMapping/UserData/blob/484ce43b2eed0fe53f265917dd2c12a824d306f3/list_files/master_son.list#L364-L365) have always been immune to the [breaks](https://forum.travelmapping.net/index.php?topic=3648.msg19036#msg19036) that route-flipping was designed to fix. 😄

> Once the file is linked, we should mention the new log file on the forum (contributor only?) to trigger the affected hwy data managers. Well... mostly Si

Or, even before it's linked. [Hey why not.](https://forum.travelmapping.net/index.php?topic=4832)

michihdeu:
> I think D1001 should have also hade the points reversed rather than the routes, to go N-S.

It is a **former** N route heading north from Paris 🤣

PR #495, opened 2022-02-18
user logs: "Please report" DISCONNECTED_ROUTE

yakra:
In the same vein as the existing "Please report" for DUPLICATE_LABEL.


yakra:
This will also be superseded.

PR #502, opened 2022-02-27
update notification bugfixes (#500) / speedup

### LINKED ISSUE #500: formerly connected routes
These code blocks appear to be indented 1 "unit" (4 spaces) too far:
https://github.com/TravelMapping/DataProcessing/blob/3cc4219090eb4bc2341571a7287442dcdfe054ed/siteupdate/python-teresco/siteupdate.py#L1602-L1607
https://github.com/TravelMapping/DataProcessing/blob/3cc4219090eb4bc2341571a7287442dcdfe054ed/siteupdate/python-teresco/siteupdate.py#L1614-L1619

C++ version has it right, I believe.
Make test cases & check for diffs:
* A61 thru Garzweiler II
* US20 thru Yellowstone

yakra:
* Closes #500
  * Fixes Python bug where entries for the first chopped route in a ConnectedRoute would prevent update notifications for the last chopped route for "not in same connected route" errors.
  * Always show updates for broken entries; avoid having good entries updated before the .list file prevent them
* Closes yakra#202
  * C++ is about 2% faster Processing traveler list files.
  * Python is faster only in theory, not noticeably so in practice. Within margin of error.

### DIFF
--- a/siteupdate/cplusplus/classes/Route/Route.cpp
+++ b/siteupdate/cplusplus/classes/Route/Route.cpp
@@ -237,7 +237,7 @@ void Route::store_traveled_segments(TravelerList* t, std::ofstream& log, unsigne
 		hs->add_clinched_by(t);
 		t->clinched_segments.insert(hs);
 	}
-	if (t->routes.insert(this).second && last_update && t->update && last_update[0] >= *t->update)
+	if (last_update && t->updated_routes.insert(this).second && t->update && last_update[0] >= *t->update)
 		log << "Route updated " << last_update[0] << ": " << readable_name() << '\n';
 }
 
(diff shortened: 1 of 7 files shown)


jteresco:
In for the next update.

yakra:
List502
I didn't initially include a chart, because the old<->new speeds for each machine are so tiny. Barely any visible distance. LOL that's a typo. Visible *difference*.

Dropping this here now because it shows how the different machines stack up against one another, and more importantly, it's one more task for which I have FreeBSD data ready to go.

Normally, the caption up top would list the HighwayData & UserData commits used in the speed tests, but they were erroneously recorded as the *DataProcessing* commits seen in the legend on the right instead. So I just removed that info.

PR #505, opened 2022-03-20
various minor improvements

yakra:
### d2cdcaab8fab8cc73561d7870186b800d1f7acb8 set/map/dict cleanup
closes yakra#199

On the C++ side, this was mostly conceived as code-readability cleanup, but it ended up saving a little time on userlog generation, more noticeable on slower machines like lab3.

Python: In theory, saves a wee bit of time, not having to create new list objects from objects that are already iterable/searchable. In practice, it's down near the margin of error; the tiny improvements we can consistently see disappear under the larger fluctuations that happen naturally between runs in other parts of the program. Either way, the code's slightly cleaner.

---

### c8b3c9829367d40c00641ee0ad0fe8cbda47ac80 eliminates an unnecessary membership test

https://github.com/TravelMapping/DataProcessing/blob/d2cdcaab8fab8cc73561d7870186b800d1f7acb8/siteupdate/python-teresco/siteupdate.py#L3603

https://github.com/TravelMapping/DataProcessing/blob/d2cdcaab8fab8cc73561d7870186b800d1f7acb8/siteupdate/cplusplus/classes/TravelerList/userlog.cpp#L68

The first half of the check is unnecessary because to get here...
https://github.com/TravelMapping/DataProcessing/blob/d2cdcaab8fab8cc73561d7870186b800d1f7acb8/siteupdate/python-teresco/siteupdate.py#L3580-L3583

https://github.com/TravelMapping/DataProcessing/blob/d2cdcaab8fab8cc73561d7870186b800d1f7acb8/siteupdate/cplusplus/classes/TravelerList/userlog.cpp#L43-L46

...`t_system_overall` has to be non-zero. But it's just been initialized to zero, and the only way for it to be set otherwise is `if h.systemname in t.system_region_mileages`.

---

### 597682e8535f1a1f08b26e42e9b6a08c5b997182 pyfloat
closes yakra#196

C++: minor speed improvements emulating "Python-style floats", where a single ".0" is appended to integer values.
* `Waypoint::str` and `Waypoint::nmplogs` write to nearmisspoints.log & tm-master.nmp respectively.
* Rather than apply similar changes to `Route::csv_line` and optionally adding "pyfloat" support to `ConnectedRoute::csv_line` to conform, these functions were removed in favor of inserting their elements directly into the file stream in sqlfile, for a small speedup. Python-style floats were dropped in the process; the odds of a Route or ConnectedRoute's mileage being an exact integer are close to zero. I'll cross that bridge if & when I come to it.
* `TravelerList::userlog` is no-build for now, to avoid merge conflicts. I anticipate refactoring the relevant bits of code into the `sqlfile1` function.

---

### 6e0693fba6462f2ef4fca142ce87882a9171e1ad fix potential buffer overflow & a3b35e045d0e319e37bd8641614b6b14c54d3531 comments
are what they say on the tin.

---

### 3ae68f11245a867ec2677424232eabbea28b244d don't store `update` in TravelerList
saves a tiny bit of RAM.

---

### d56e5a67555922d06730363ca779c0acdc224dac update info retrieval
cleans up & consolidates the code that finds the update date & time for each .list file upon construction of its TravelerList object.

### DIFF
--- a/siteupdate/cplusplus/classes/ConnectedRoute/ConnectedRoute.cpp
+++ b/siteupdate/cplusplus/classes/ConnectedRoute/ConnectedRoute.cpp
@@ -76,14 +76,6 @@ std::string ConnectedRoute::connected_rtes_line()
 	return line;
 }
 
-std::string ConnectedRoute::csv_line()
-{	/* return csv line to insert into a table */
-	char fstr[32];
-	sprintf(fstr, "','%.15g'", mileage);
-	return "'" + system->systemname + "','" + route + "','" + banner + "','" + double_quotes(groupname)
-		   + "','" + (roots.size() ? roots[0]->root.data() : "ERROR_NO_ROOTS") + fstr;
-}
-
 std::string ConnectedRoute::readable_name()
 {	/* return a string for a human-readable connected route name */
 	std::string ans = route + banner;
--- a/siteupdate/cplusplus/classes/ConnectedRoute/ConnectedRoute.h
+++ b/siteupdate/cplusplus/classes/ConnectedRoute/ConnectedRoute.h
@@ -21,7 +21,6 @@ class ConnectedRoute
 	ConnectedRoute(std::string &, HighwaySystem *, ErrorList &);
 
 	std::string connected_rtes_line();
-	std::string csv_line();
 	std::string readable_name();
 	//std::string list_lines(int, int, std::string, size_t);
 };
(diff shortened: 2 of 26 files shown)


yakra:
I thought this was ready for production, but at the last minute have come across some even better ways to approach some of the things that have been changed. Putting this on hold.

PR #507, opened 2022-03-22
various minor improvements, take 2

yakra:
The first 5 commits are rebased from the last 5 in #505.
* 571e1360afe5c430ae7762d0a04191fe9a543bcc closes yakra#196. Refactors DB lines for Routes & ConnectedRoutes, and improves writing "Python-style floats", with a single `.0` appended to integer values, to nearmisspoints.log & tm-master.nmp.
* 58b4b5737913ec7b8463620c4d7017d89724eb9c & 2e6dc1a7a6a69267ab8fdd15b7889db69fcbd69e are what they say on the tin.
* 26f547879af8c0b95208b66fd518e8f00ddc3b21 saves a tiny bit of RAM for each `TravelerList`.
* 5b68177731ee46ba41791db41a2fef295a67af75 refactors finding the update date & time for each .list file upon construction of its TravelerList object.

The last 2 commits are redos of the first 2 in #505.
[There are several ways to more effectively use C++'s `std::unordered_map` class.](https://github.com/yakra/DataProcessing/issues/216) Changes that would have conflicted with the first 2 commits of #505 are redone here; other items can be explored in the future.
* 16721240fb5c3e117b1701b62bfcda6a10fb0c61 has the same diffs as d2cdcaab8fab8cc73561d7870186b800d1f7acb8 for 10 of 13 files, saving the other 3 for the next commit:
  * C++: mostly conceived as code-readability cleanup, but it ended up saving a little time on userlog generation, more noticeable on slower machines like lab3. Closes yakra#199.
  * Python: Minor cleanup that avoids having to create new list objects from objects that are already iterable/searchable.
* dd274f24be9bd0b61d47b5e4048ad0db21db6f6f redoes HighwaySystem.cpp, Route.cpp & sql_file.cpp.
  Both before & after [#505](https://github.com/TravelMapping/DataProcessing/pull/505), time was wasted by searching hash tables twice -- once to determine membership and again to get the value for a given key.
Before, we at least *got* the result of `std::unordered_map::find`, even if only to use it as a temporary & then just discard it.
Here, that result gets saved & used a 2nd time, avoiding that 2nd lookup.
**Plus,** we get rid of that unnecessary membership test as done in c8b3c9829367d40c00641ee0ad0fe8cbda47ac80, described with the code snippets in [#505](https://github.com/TravelMapping/DataProcessing/pull/505)'s OP. Not its own separate commit this time around.

### DIFF
--- a/siteupdate/cplusplus/classes/ConnectedRoute/ConnectedRoute.cpp
+++ b/siteupdate/cplusplus/classes/ConnectedRoute/ConnectedRoute.cpp
@@ -76,14 +76,6 @@ std::string ConnectedRoute::connected_rtes_line()
 	return line;
 }
 
-std::string ConnectedRoute::csv_line()
-{	/* return csv line to insert into a table */
-	char fstr[32];
-	sprintf(fstr, "','%.15g'", mileage);
-	return "'" + system->systemname + "','" + route + "','" + banner + "','" + double_quotes(groupname)
-		   + "','" + (roots.size() ? roots[0]->root.data() : "ERROR_NO_ROOTS") + fstr;
-}
-
 std::string ConnectedRoute::readable_name()
 {	/* return a string for a human-readable connected route name */
 	std::string ans = route + banner;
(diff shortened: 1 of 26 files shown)


jteresco:
all good to go to production?

yakra:
Good to go.

PR #508, opened 2022-03-25
NMP speed improvements

### LINKED ISSUE #457: Single-threaded NMP detection: restore Jim's original formula
https://github.com/TravelMapping/DataProcessing/blob/082e9938ed5a2ddd39985f570cf34080abfb76ef/siteupdate/cplusplus/siteupdate.cpp#L258
* This wastes time iterating thru quadtree nodes & constructing the `point_list`, and then iterating thru *that*.
* Iterating thru `HighwaySystem`->`Route`->`Waypoint` would mean iterating thru slightly more containers to get at our points, but save us the cost of list construction. Might *(probably?)* be faster, might be slower.
* An even better solution would be to restore Jim's original method to siteupdateST via conditional compilation. NMP detection [was moved](https://github.com/TravelMapping/DataProcessing/pull/280/files#diff-2f5473880a30bc31196285e141743f822d5016cab7c4e72fb95946874519f63c) in #280 to benefit multi-threaded performance, with the side effect of hurting single-threaded performance (charts are at #284).
* https://github.com/TravelMapping/DataProcessing/issues/141#issuecomment-569603359
  > At 1 thread, we start off a smidgen less efficient, surely due to the reasons [noted upthread](https://github.com/TravelMapping/DataProcessing/issues/141#issuecomment-444777451).
At 2 threads, we're off and running, with the gains from multithreading outweighing the loss in efficiency in the raw NMP search.
* Those reasons being:
  * https://github.com/TravelMapping/DataProcessing/issues/141#issuecomment-444777451
    > Checking for NMPs after the quadtree is completely populated means comparing every point A against every point B, point C, etc. and vice versa within the terminal nodes encompassing `tolerance`.
    >
    > As Jim has it set up now...
    > 1. Point A is added to the quadtree first. There will be no other NMPs to check against.
    > 2. Point B is added 2nd, and point A is detected. A is added to B's list, and B to A's. No checking against point C which has not been added yet.
    > 3. Point C added, A & B detected, yadda yadda.
    >
    > Faster.
  * https://github.com/TravelMapping/DataProcessing/issues/141#issuecomment-444780792
    > When earlier points are added and checked, fewer points in the quadtree means iterating down thru fewer refined nodes.
Over half a million points, the savings add up.

See also: #393

yakra:
## C++
Two components here:
* Close yakra#157
  Don't create a bunch of temporaries and muck about with move constructors & splicing; just push points directly to the nmp list in the `Waypoint` object itself. Saves a rather surprising amount of time for what it is.
  nmpdirect
* Close #457
  When #280 were enabled, multi-threaded NMP searching came at the cost of a performance hit in single-threaded searches. Background is at #141.
  Conditioning Jim's original formula back in for single-threaded siteupdateST is easy.
  **Speed comparison** after the above component, constructing in place:
  * Time no longer spent specifically `Searching for near-miss points` via **Eric's** method should outweigh
  * the additional time `Reading waypoints for all routes`, which represents **Jim's** method.
    Box | Eric’s | Jim’s
    -- | -- | --
    BT | 1.72 | 1.437
    lab1 | 1.009 | 0.6198
    lab5 | 1.435 | 1.033
    lab2 | 1.349 | 0.966
    lab3 | 1.79 | 1.092
    lab4 | 2.16 | 1.601
    bsdlab | 2.581 | 1.752
* A third component I tried out & rejected was to
https://github.com/TravelMapping/DataProcessing/blob/de796fa866faa446f8e2d098b87ba11a959128e9/siteupdate/cplusplus/classes/Waypoint/Waypoint.cpp#L219 in `NmpSearchThread` rather than a single-threaded part of the program. Turns out this is slower.
What can I come up with for an explanation?
  * Turning cache hits into cache misses -- the single-threaded bit already deals with `root` & `label`.
  * Branch prediction effects, whether sorting is done at the same time as iterating down through the quadtree.

---

## Python?
Python OTOH has kept Jim's original "search a partially populated quadtree while reading all waypoints" NMP method the whole time. The only thing to do here is to try out building the NMP lists directly in the `Waypoint` object.
We're still creating and copying (not moving, I presume?) from a bunch of temporaries, right? I'd expect better results than:
Box | de796fa | a768800
-- | -- | --
epoch | 115.24 | 120.19
BT | 113.3 | 111.59
lab1 | 56.98 | 57.07
lab5 | 51.18 | 51
lab2 | 72.24 | 72.09
lab3 | 98.15 | 97.2
lab4 | 80.26 | 79.61
bsdlab | 126.58 | 125.06
noreaster | 93.46 | 92.53

Python is pretty mysterious under the hood. Maybe it has to check whether `w.near_miss_points` has an `append` method every time when attempting to call it, and OTOH knows that locally-declared `near_miss_points` is a list? I don't know man, I didn't do it.

I'm undecided about whether to include this. Sometimes I like to make Python changes that mirror C++ changes, just to keep the two versions structured similarly. Other times I'll leave out changes with no clear benefit, to have a smaller impact on the commit history. Is there a clear benefit here? \*Shrug.\* At least we're down 14 lines of code?
The diff is at a7688007907ebf3fe9a6a16b0b8c8825bd693cfb. @jteresco, say the word and I can push it to this branch. Or leave it out.

### DIFF
--- a/siteupdate/cplusplus/Makefile
+++ b/siteupdate/cplusplus/Makefile
@@ -3,12 +3,14 @@ CXXFLAGS = -std=c++11 -Wno-comment -Wno-dangling-else -Wno-logical-op-parenthese
 
 MTObjects = siteupdateMT.o functions/sql_fileMT.o threads/threads.o \
   classes/GraphGeneration/HighwayGraphMT.o \
+  classes/Route/read_wptMT.o \
   classes/WaypointQuadtree/WaypointQuadtreeMT.o \
   functions/allbyregionactiveonlyMT.o \
   functions/allbyregionactivepreviewMT.o
 
 STObjects = siteupdateST.o functions/sql_fileST.o \
   classes/GraphGeneration/HighwayGraphST.o \
+  classes/Route/read_wptST.o \
   classes/WaypointQuadtree/WaypointQuadtreeST.o \
   functions/allbyregionactiveonlyST.o \
   functions/allbyregionactivepreviewST.o
@@ -30,7 +32,6 @@ CommonObjects = \
   classes/HighwaySystem/route_integrity.o \
   classes/Region/Region.o \
   classes/Route/Route.o \
-  classes/Route/read_wpt.o \
   classes/TravelerList/TravelerList.o \
   classes/TravelerList/userlog.o \
   classes/Waypoint/Waypoint.o \
--- a/siteupdate/cplusplus/classes/Route/read_wpt.cpp
+++ b/siteupdate/cplusplus/classes/Route/read_wpt.cpp
@@ -65,11 +65,19 @@ void Route::read_wpt(WaypointQuadtree *all_waypoints, ErrorList *el, bool usa_fl
 		if (!*lines[l]) continue;		// whitespace-only line; skip
 		Waypoint *w = new Waypoint(lines[l], this);
 			      // deleted by WaypointQuadtree::final_report, or immediately below if invalid
+		// lat & lng both equal to 0 marks a point as invalid
 		if (!w->lat && !w->lng)
 		{	delete w;
 			continue;
 		}
 		point_list.push_back(w);
+
+	      #ifndef threading_enabled
+		// look for near-miss points (before we add this one in)
+		all_waypoints->near_miss_waypoints(w, 0.0005);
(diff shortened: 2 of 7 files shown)


yakra:
Sounds good. Let's save those 14 lines of code, and that 0.93s when running datacheck.sh!
Pushed a768800 to this branch.

jteresco:
This will be used for tonight's update.

PR #512, opened 2022-04-01
eliminate clinchedDBValues class

### LINKED ISSUE #511: sprintf precision
C++: `routes` and `connectedRoutes` are written to the DB with 17 & 15 sig figs respectively.
Meanwhile, Python seems to keep a pretty consistent 16 with its `str(double)`.

Could `%.16g` be the magic bullet?
**Edit:** Nope. `%.17g` then?
```sh
egrep -c '[1-9][0-9.]{17}' ~/ytm/DataProcessing/siteupdate/python-teresco/TravelMapping.sql
414823
```
...Nope again. Lat / long coords want 15 sig figs. Any more and we can get lots of 0s followed by another digit appended after the _real_ value.
Looks like I was on the right track originally - 15 sig figs in some cases, 17 in others.
* [x] Try 15 in only lat/long cases, 17 in all others

---

* [x] clinchedConnectedRoutes & clinchedRoutes also need one more sig fig.
  This can get squashed into the ongoing changes on the "r4" branch.
* [x] clinchedSystemMileageByRegion too, in theory.
  Would it help in practice? (No. But anyway.)

yakra:
## C++
Closes yakra#161. Instead of getting into the wackier ideas proposed there, `cr_values` and `ccr_values` are stored as pointer/double pairs inside each `TravelerList`, eliminating the need for mutexes without having to resort to arrays of containers. Got rid of `csmbr_values`; its lines are generated on-the-fly as the .sql file is written. This lets us scrap the `ClinchedDBValues` class entirely, a win for code organization.

This started out by looking at that data structure, saying "Hey, this stores a whole lot of text that just *hangs out* until the DB is written. What can we do about this?," and then thinking about how all those string allocations, re-allocations & copies, along with the use of a mutex, must get expensive.

### Speed
Time savings of about 4% with fewer threads on Linux, up to 10%, 11%, even 15% at the high end on lab3 & lab4.
User-512
And then there's FreeBSD, which is a lot more restricted by the RAM bottleneck. Big win here!

The savings in *Creating per-traveler stats log entries and augmenting data structure* lead to more complexity when writing the .sql file. A task that on its own is single-threaded, which would eventually mean more total time at more threads, except we make up for that by running it in the background during subgraph generation. The extra time is thus divided by `numthreads`, roughly speaking.
On FreeBSD, the DB tables go from 0.1s to 1.2s, with the much larger `UserLogThread` savings alone making up for it at any `numthreads`.

### RAM
I estimate we're saving 60.4 MB on Ubuntu, 57.4 MB on FreeBSD.
System Monitor indicates between ~0-24 MB, though that's not the most precise tool.

### .sql file
The DB structure itself doesn't change, but the .sql file has some minor differences.
* **Before:** `clinched(Connected)Routes` values were grouped into sets of 10,000 per INSERT command.
* **After:** For expediency's sake, repurposed the existing outer/inner loops to iterate through travelers, then their respective pointer/mileage pairs.

With 322 travelers, This means 1322 insertions/query for clinchedRoutes & 1096 insertions/query for clinchedConnectedRoutes.
[We talked a bit about optimizing the number of inserts per query a while back](https://forum.travelmapping.net/index.php?topic=3102.msg14324#msg14324).
I split up the .sql file into smaller files, and fed them into MySQL individually:
* **file1.sql**, everything before clinchedConnectedRoutes, took 2:54 to ingest the 1st time & 2:51 the 2nd.
  So there's an indication of what kind of natural speed variations we might see.
* **file2a.sql**, clinchedConnectedRoutes & clinchedRoutes produced by 1279af8, 10000 inserts/query, took 0:20 to ingest.
* **file2b.sql**, clinchedConnectedRoutes & clinchedRoutes produced by 032c96d, 1209 inserts/query, took 0:18 to ingest. Significant?
* **file3.sql**, everything after clinchedRoutes, took 0.03 both times.

Closes #511.
Tweaks the sig figs written to the .sql file in the C++ version.
C++ now uses "Python-style floats" with a trailing `.0` on integers in the DUPLICATE_COORDS & OUT_OF_BOUNDS datachecks.

## Python?
This is another one I'm on the fence about including. Diff is at b914292028b24fd8effa7528cd4b6ba3e651651a; I can push it to this branch.
The changes are less help with speed here, just taking the time spent in one part of the program and moving it elsewhere.
I blame the file `write` function, as it needs one string to write, which must be allocated & constructed & copied & whatever goes on under the hood. *(When I've tried using multiple `write`s to get around this, things get *slower*.)*
That and the fact that we're now iterating through multiple nested containers & checking for empty containers.

* **Speed**
  * The entire siteupdate program is slower on average 0.67 s across the six machines I tested.
  * For those just running in errorcheck mode & not writing the DB file, it's faster 2.4 s on average.
* **RAM**
  * System Monitor suggests savings of ~8-40 MB, more than what I observed for C++, but again, not the best yardstick.
  * I expect bigger RAM savings than the C++ version, which again should be in the 60 MB ballpark.
    Comparatively smaller for the gobs of RAM Python uses, but still.
* **.sql file**
  * While C++-generated & Python-generated files won't be 100% diff-able, keeping the same number of lines (of SQL source) per table seems like a good thing.
* **Code org.**
  * This keeps the C++ & Python versions a tiny bit less divergent. Similar changes in both.
  * Down 8 whole lines of code!
* **Con:**
  * Slower overall by 0.67 s.
* **Pro:**
  * Faster 2.4 s in errorcheck mode.
  * Smaller RAM footprint. (Slightly?)
  * Keep program structure more similar.
  * Keep DB file structure more similar.
  * Down 8 lines of code.

@jteresco, thoughts?


yakra:
Let's wait a tick on merging this in.
I've come across some more efficiency improvements I can make to the userlog routines.
Getting rid of `csmbr_values` is a prerequisite. If this works out well, it'll make including the Python changes a foregone conclusion. :)

PR #524, opened 2022-04-23
region management / remove redundant computation

### LINKED ISSUE #522: redundant active/preview system counts
https://github.com/TravelMapping/DataProcessing/blob/9dcf0a5549cc6100082710637e2653e99afd8c80/siteupdate/cplusplus/classes/TravelerList/userlog.cpp#L32-L41
https://github.com/TravelMapping/DataProcessing/blob/cc9da66106a7581974938040418b845dcedaf4b2/siteupdate/python-teresco/siteupdate.py#L3543-L3554
These values don't need to be computed for every user log. Compute once, while
* writing highwaydatastats.log, or
* reading systems.csv *<--probably this?*
* or like whatever

While in there, should also remove the comment, as `csmbr_values` is no longer a thing as of 9dcf0a5549cc6100082710637e2653e99afd8c80.

### LINKED ISSUE #523: redundant active/preview mileage
Active/preview mileage is saved in variables when writing highwaydatastats.log.
No need to recalculate for allbyregionactiveonly.csv & allbyregionactivepreview.csv; just:
* C++: pass the appropriate variable in
* Python: reuse the appropriate variable

yakra:
Closes yakra#220. Removes an unused vestigial variable.
Closes #523.
Closes #522.

---

**C++ Region management:**
* A step toward making siteupdate "Valgrind clean". Rather than manually create `new` Region objects on the heap and store pointers in a `vector`, Regions are emplaced in a `list`. *(I'd prefer a `vector`, but can't because `Region` contains a `std::mutex`, which is immovable, making `Region`s immovable too. Went with this instead.)*
I don't have to manually delete anything, as the list destructor cleans the objects up at the end of the program when the list goes out of scope.
*Next up: TravelerLists.*
* In the process, changed the `Region` ctor to throw if the regions.csv line has the wrong # of fields. AIUI that's considered better practice than setting & checking a `valid` bit and deleting a fully constructed object, popping it from a container, etc. Cleaner code either way, and a tad less RAM.

### DIFF
--- a/siteupdate/cplusplus/classes/HighwaySystem/HighwaySystem.cpp
+++ b/siteupdate/cplusplus/classes/HighwaySystem/HighwaySystem.cpp
@@ -12,6 +12,8 @@
 
 std::list<HighwaySystem*> HighwaySystem::syslist;
 std::list<HighwaySystem*>::iterator HighwaySystem::it;
+unsigned int HighwaySystem::num_active  = 0;
+unsigned int HighwaySystem::num_preview = 0;
 
 HighwaySystem::HighwaySystem(std::string &line, ErrorList &el, std::vector<std::pair<std::string,std::string>> &countries)
 {	std::ifstream file;
@@ -54,7 +56,10 @@ HighwaySystem::HighwaySystem(std::string &line, ErrorList &el, std::vector<std::
 	level = level_str[0];
 	if (level_str != "active" && level_str != "preview" && level_str != "devel")
 		el.add_error("Unrecognized level in " + Args::systemsfile + " line: " + line);
-
+	switch(level)
+	{	case 'p': num_preview++; break;
+		case 'a': num_active++;
+	}
 	std::cout << systemname << '.' << std::flush;
 
 	// read chopped routes CSV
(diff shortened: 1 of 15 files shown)


yakra:
Going to take a closer look at this one. May reopen in a few... hours? Days?

yakra:
Expected to re-open this after double-checking all my silly benchmarks.
Around the time I closed it, I found a couple things to optimize in `TravelerList::userlog`. One (counterintuitively) didn't pan out, but the other did. A decent performance boost under FreeBSD -- not as much as I'd hoped, but an improvement nonetheless.

Meanwhile, I've hit upon a way to get Region objects into contiguous memory, mutexes and everything. All I've got to do is:
* Refactor the code that reads .wpt & .list files out into a new reusable function. *(Done!)*
* Write a new .csv parsing routine. *(Mostly done, as a separate program not integrated into siteupdate yet.)*
* Change the Region constructor to take fields as args instead of a whole line. *(To do.)*
* Write a custom container class. *(To do.)*

Easy, right? :rofl: This will pay off when I get around to arranging HighwaySystems, Routes & HighwaySegments contiguously.

yakra:
> A decent performance boost under FreeBSD -- not as much as I'd hoped, but an improvement nonetheless.

Found a way to take it a step farther. Be a few hours before I can test on bsdlab (just started BiggaTomato) but the concept looks promising.

PR #520, opened 2022-04-18
use clang optimization flag

yakra:
Be sure to `make clean` before you `make` again (or `gmake` as the case may be).

Re the change to WaypointQuadtree.cpp, I'm surprised I never got linking errors before.

---
RWpt
Sort
Wow, FreeBSD is a top performer here. Note the small bump in processing time when we run out of physical cores & start using hyperthreaded virtual cores.
Less apparent on lab3 & especially lab4.
NmpS
The "default" line for FreeBSD was thrown off by a couple big outliers.
NmpM
Ouch, [another horrible performer under FreeBSD](https://github.com/TravelMapping/DataProcessing/issues/519).
NmpM_Linux
Same thing, zoomed in to just the Linux machines.
Intg
Another win for FreeBSD.
List
Conc
A couple outliers in FreeBSD's "default" line again.
csr
For the 24-thread machines, the results of 5 passes were averaged together instead of the usual 10, because I wanted to do these tests in a finite amount of time. :) With this task being so fast, the noise is a much larger piece of the pie, resulting in fairly jagged lines.
cst
User
Another case where FreeBSD has trouble scaling beyond just a few threads, though this isn't as bad as the others.
There are several potential changes that could maybe help out, and as we saw in https://github.com/TravelMapping/DataProcessing/pull/517, sometimes a little tweak goes a long way.
Good news is, noreaster scales up to 8 threads. This looks like a memory bandwidth issue, and shouldn't be as tough to tackle as nmp_merged or subgraphs.
Vert
The somewhat jagged lines could be a result of small sample size again.
Why does BiggaTomato stay flat between 3 & 4 threads? The DB file is being written in the background. So for a little while, siteupdate actually runs `numthreads`+1 threads. It's just easiest to do this way. It usually finishes up at some point during subgraph generation.
Well, that _would_ be the case, except that it already stays flat between 2 & 3 threads. Running out of memory bandwidth I guess.
Subg
This is just painful. :( A nice bump at 1 thread, and then... nothing. [This may take a lot of work to see an improvement...](https://github.com/TravelMapping/DataProcessing/issues/519)
Subg_Linux
Same thing, zoomed in to just the Linux machines.
At 4 threads on BT & 8 on lab1 & lab5, that's probably the DB file slowing down graph generation again.
mtsub
Total

---

### single-threaded tasks in multi-threaded `siteupdate`
**top:** default / `-O0`
**bottom:** `-O3`

|BT|lab1|lab5|lab2|lab3|lab4|bsdlab|
[...]

PR #515, opened 2022-04-12
graph generation speedups

yakra:
@jteresco, this doesn't make any of the changes I PMed you about on the forum.
What *is* here is mostly C++, various graph generation optimizations. The ones that I could also do in Python I did there too, though the results were underwhelming.

---

6c97b69059ba49775080f367416d501251e3e282 closes yakra#195.
Area graph vertex sets are populated in place instead of dealing with a bunch of temporaries, same idea as what #508 did with NMPs.
* C++: cuts area graph time by ~28%. Though area graphs _already_ took only 0.49-0.79 s depending on machine.
* Python: saves about 1.5%.

---

33183f2d8bb2f85617b30ad147fce252fb389305 takes advantage of memory locality (and probably also faster iteration?) and speeds up:
* initial vertex creation
* getting matching sets for subgraphs

[Some harmless redundancy is involved.](https://forum.travelmapping.net/index.php?topic=2926.msg27248#msg27248)
Writing subgraphs, saves ~5-7% depending on machine @ 1 thread, tapering off as thread count increases. Hm. Wonder why that is. *Cache line bouncing?*

---

f0d618ae25fd60c6ca598907eaf8c4ab7ca37f9a is a step toward the long-term goal of moving as much code as practical out of the main siteupdate.cpp file, whether by function calls or simple `#include`s.

---

12beccd5c04b56d12bc11ba88da0930d35d9212b uses a `switch` when writing subgraph vertices. Much more concise. Will help with readability when https://github.com/TravelMapping/EduTools/issues/156 happens.

---

f699eba4c9f92eaccd01c120141e50c0a134de50 is 5 commits squashed together. Now we're getting into the good stuff!
* **V1a:** store vertices in contiguous memory
  Writing subgraphs, saves 5-9% @ 1 thread on the various Linux machines, tapering off as thread count increases.
  This sets the stage for [the bit field I mentioned on the forum](https://forum.travelmapping.net/index.php?topic=2926.msg27239#msg27239), but doesn't implement it just yet.
* **V1b:** `switch`es when writing master graph vertices & edges
* **T:** allocate `clinchedby_code` once per graph; modify in place
  This avoids having to allocate a string for every edge, instead reusing one `char*` for each edge in a graph.
  * Ubuntu: Saves ~9.5% on subgraphs, looking fairly flat WRT thread count.
[...]

PR #526, opened 2022-04-27
user log speedup / remove redundant computation

yakra:
This is a do-over of #524.

Closes https://github.com/yakra/DataProcessing/issues/220.
Closes https://github.com/TravelMapping/DataProcessing/issues/523.
Closes https://github.com/TravelMapping/DataProcessing/issues/522.

---

**C++ Region management** is on the backburner.
There are a couple different paths forward here, and I'm not 100% sold on either. Kicking that can down the road.

---

### User log speedup
Small improvement on Linux; big one for FreeBSD.
For just FreeBSD, the chart below includes an intermediate step:
the initial conversion of `std::string to_write` to `std::vector<std::pair<Route*, double>> chop_mi`
(not included for Linux, as the lines are already so close together it'd just add clutter).
After that, the next step was to have `format_clinched_mi` format & return a `char*`. Same idea as what #515 did with `clinchedby_code` in traveled graphs.
User526-18
[Alternate version](https://user-images.githubusercontent.com/13720877/165646041-ab69964b-91fb-4382-a1b9-0c9aa07014c6.png) with 2x the vertical scale, to compare against [this one](https://user-images.githubusercontent.com/13720877/163829580-9c9b2d2c-0299-45a2-a023-d4093a3c67bf.gif) from #520. And before that, there was #513.

**Future?**
There may be more improvements to be had by playing around with the `std::unordered_map`s or converting to `std::map`s.
Sometimes small changes make big results (see #517). But OTOH, we could be approaching the point of diminishing returns.
Call me cautiously optimistic.

### DIFF
--- a/siteupdate/cplusplus/classes/HighwaySystem/HighwaySystem.cpp
+++ b/siteupdate/cplusplus/classes/HighwaySystem/HighwaySystem.cpp
@@ -12,6 +12,8 @@
 
 std::list<HighwaySystem*> HighwaySystem::syslist;
 std::list<HighwaySystem*>::iterator HighwaySystem::it;
+unsigned int HighwaySystem::num_active  = 0;
+unsigned int HighwaySystem::num_preview = 0;
 
 HighwaySystem::HighwaySystem(std::string &line, ErrorList &el, std::vector<std::pair<std::string,std::string>> &countries)
 {	std::ifstream file;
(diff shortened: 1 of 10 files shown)


yakra:
Doing one last comparison.

yakra:
Good to go.

jteresco:
Bringing this in for the site update that I'm about to launch.  Thanks.

PR #534, opened 2022-05-30
nmpbyregion 2.0

### LINKED ISSUE #533: Site update: NMP by region page improvements
https://forum.travelmapping.net/index.php?topic=3656.msg27805;topicseen#new

Two possibilities:

1. Just don't generate `.nmp` files for regions with no NMPs
2. Don't generate entries on the `index.html` file for 0-byte `.nmp`s.

yakra:
closes #533

I decided to order the entries by descending number of unmarked entries, as it was simplest to code.

There are no longer any hyphens in filenames -- `nmpbyregion` only uses info from tm-master.nmp itself now, so we only have lower-case regions with no hyphens, as used in .wpt file roots, available.
The old way of searching `hwy_data/` for region codes & converting them was bloaty overkill that contributed to all the 0-byte files. I figure converting `deurp` to `DEU-RP` etc. is more trouble than it's worth.

### DIFF
--- a/nmpfilter/.gitignore
+++ b/nmpfilter/.gitignore
@@ -1,0 +1,1 @@
+nmpbyregion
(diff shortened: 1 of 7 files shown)


jteresco:
Thanks!  I'll try this out with tonight's site update.

PR #538, opened 2022-07-10
PlaceRadius::r -> double

### LINKED ISSUE #536: Allow non-integer radius for area graphs in site update
I've updated the Python site update code to allow non-integer values for the area graphs.  Will need the same in C++ if it's not already allowing that.

yakra:
closes #536


jteresco:
Thanks for this.

PR #541, opened 2022-10-03
avoid crashes due to 0-byte .list files

yakra:
https://github.com/TravelMapping/UserData/pull/7658#issuecomment-1265865478


yakra:
246f22b379207afe347d35d28c24aaf62f765bcd
Know what, I think I see a cleaner way to do this. Hang on.

yakra:
Force pushed. Ready to rock.

jteresco:
Thanks - trying it out in tonight's update.

PR #547, opened 2022-12-28
LABEL_LOWERCASE & LOWERCASE_SUFFIX (DataProcessing)

### LINKED ISSUE #488: Waypoint labels (or label suffix) beginning with lowercase character
Just a little challenge for @yakra 

1. Do we have labels starting with lowercase characters?
2. Are they incorrect or do we have FPs?
2. Is it worth a data check?

It sometimes happens to me that I create wpts like rueMar instead of RueMar. Unfortunately, the wpt editor doesn't flag it wrong.

yakra:
closes #488


yakra:
That was quick! Don't forget https://github.com/TravelMapping/Web/pull/739 :)

PR #558, opened 2023-01-07
.list update time via .time file

### LINKED ISSUE #550: Efficiency of creating listupdates.txt
The site update scripts create a file `listupdates.txt` that is used by the Python and C++ site update programs to get the timestamp of the last commit that modified each list file.  It's a slow process, with a call to `git log` for each list file.

Could this be improved by having a `Makefile` that creates something like a `terescoj.time` file with this info for each list entry, and which would only then need to make the `git log` call when the `.list` file has been updated?

yakra:
Along with https://github.com/TravelMapping/UserData/pull/8069,  this closes #550.

### DIFF
--- a/siteupdate/cplusplus/classes/Route/Route.cpp
+++ b/siteupdate/cplusplus/classes/Route/Route.cpp
@@ -215,14 +215,14 @@ void Route::write_nmp_merged()
 	wptfile.close();
 }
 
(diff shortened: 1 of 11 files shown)


jteresco (approved):
All looks reasonable to me, thanks!  I'll try it with tonight's update.

PR #560, opened 2023-01-07
insert into graphs immediately upon creation

yakra:
https://github.com/TravelMapping/DataProcessing/commit/2c541b1bdf30316dae13fc8a50c8e8a6ff95a613#commitcomment-94744964
> 2. `compare_sql` still does its job, but due to the way the tables are set up misreports the number of lines in a few tables. This is because it expects a table's insertions to be made immediately after it's created. Instead, the `INSERT INTO graphs VALUES` lines appear after `graphArchiveSets` and `graphArchives` are created. Thus `graphs` is incorrectly reported as 1 line, and `graphArchives` is incorrectly reported as 1284 lines, the number of lines in `graphs`.

> 4. I would however like to move the new `graphs` insertions back to immediately follow table creation to get `compare_sql` to report the number of lines correctly, in both C++ & Python. Any objections?


jteresco:
I will take a look this weekend, thanks.

Issue #466, opened 2021-10-02
lines[l], orig_line, trim_line


yakra:
Leaving this here because I keep asking myself, and forgetting, why we need all 3:
* `lines[l]`, the *original* `char*` buffer, is modified in place by splitting into fields.
* `orig_line` is written unmodified to split-region .list files, to minimize diffs.
* `trim_line` has whitespace stripped, and control chars are replaced with `?`s before writing notes to the userlog.

yakra:
## `fields`:
`char*` -> `std::string` conversions are performed for...
* hash lookups
* set lookups
* set insertions
* set erasures
* creating `lookup` keys (see also yakra#191)

This all adds up.
If `fields` were created as `std::string`s from the get-go, it could
* save some time on its own
* potentially free up the original `char*` buffer to be used in place of `orig_line` or `trim_line`

yakra:
A rework here may help FreeBSD performance.
FreeBSD appears more sensitive to RAM bottlenecks. [No longer creating all the ClinchedDBValues strings improved performance considerably when `Creating per-traveler stats logs and augmenting data structure`](https://github.com/TravelMapping/DataProcessing/pull/513).
Even though this involves a different task, [`Processing traveler list files`, which does scale pretty nicely to 24 threads, but lags behind CentOS & Ubuntu a bit](https://github.com/TravelMapping/DataProcessing/pull/502).

yakra:
C++14 heterogeneous lookup

yakra:
We're finished with `fields` at the end of `mark_chopped_route_segments.cpp` and `mark_connected_route_segments.cpp`.
* [ ] Erase from the unused sets first.
* [ ] Then use `std::move` to insert into the in-use sets.
* [ ] While in there, why not https://github.com/yakra/DataProcessing/issues/210?

PR #561, opened 2023-01-07
minor userlog tweaks

yakra:
* **`system_region_miles` -> `system_miles`**
  This function was created in the early "translate from Python" days (in place of a math.fsum) and hastily named without a clear understanding of what it was doing. There's nothing *regional* about it; it's the total across *all* regions. The new name is a more accurate description.
* **Remove extraneous conditional at beginning of stats by system**
  If the system is in `system_region_mileages`, it's a virtual guarantee mileage is >0.
  To have zero mileage, we'd have to have a zero-length segment, via a BAD_ANGLE error (which we've not seen in a long time, and those are a `Datacheck::always_error` & should in theory be fixed quickly), *and* a user would have to mark only zero-length segment(s) -- using consecutive same-coords points -- in their .list file. In the unlikely event this does happen, the worst that happens is some extraneous info (a load of 0s) in that user's .log file.
* **Remove extraneous condition from updated routes list population.**
  We know `r->last_update` is true; this was a condition of items' original insertion.
  * While at it, I made the Python version more Pythonic.
  * In C++, the new list is populated during construction.

---

In other news, I've been trying out various tweaks to increase [userlog efficiency](https://forum.travelmapping.net/index.php?topic=5349.msg30150#msg30150) and hopefully help out its poor scaling under FreeBSD. Individually, each one fails to pan out except for maybe producing the tiniest little differences on BiggaTomato. Maybe if I:
* throw [everything I can think of](https://github.com/yakra/DataProcessing/issues/234) together at once, or
* find the machine with the right combo of most cores & least cache _(lab1, lab5, BiggaTomato?)_ and cripple it with a single channel of the slowest RAM I can find,

... I might see some results?

### DIFF
--- a/siteupdate/cplusplus/classes/HighwaySystem/HighwaySystem.cpp
+++ b/siteupdate/cplusplus/classes/HighwaySystem/HighwaySystem.cpp
@@ -171,7 +171,7 @@ void HighwaySystem::stats_csv()
 	{	// only include entries for travelers who have any mileage in system
 		auto it = t->system_region_mileages.find(this);
 		if (it != t->system_region_mileages.end())
-		{	sprintf(fstr, ",%.2f", t->system_region_miles(this));
+		{	sprintf(fstr, ",%.2f", t->system_miles(this));
 			sysfile << t->traveler_name << fstr;
 			for (Region *region : regions)
 			  try {	sprintf(fstr, ",%.2f", it->second.at(region));
--- a/siteupdate/cplusplus/classes/TravelerList/TravelerList.cpp
+++ b/siteupdate/cplusplus/classes/TravelerList/TravelerList.cpp
@@ -157,7 +157,7 @@ double TravelerList::active_preview_miles()
 }
 
 /* Return mileage across all regions for a specified system */
-double TravelerList::system_region_miles(HighwaySystem *h)
+double TravelerList::system_miles(HighwaySystem *h)
 {	double mi = 0;
 	for (std::pair<Region* const, double>& rm : system_region_mileages.at(h)) mi += rm.second;
 	return mi;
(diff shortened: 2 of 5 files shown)


jteresco:
Thanks for all of these and the ongoing work on efficiency overall.

yakra:
Haha, welcome! For yakra#234...
* This checks off the first 3 outer check boxes.
* Trying out the next 4 locally. More results in a few hours, but I'm expecting very little at this point. For the results I do have, it sadly looks like they're not much benefit. A bit surprisingly, because...
  * Searching an `unordered_map` once, saving an iterator & dereferencing it should be faster than searching repeatedly. Granted, these maps are small as most systems have few regions (often just 1), but still.
  A dereference *should be* faster.
  * Eliminating `HighwaySystem::total_mileage()` in favor of a variable saves us having to iterate through an `unordered_map` (which has to be more complex than iterating through a `vector`, maybe even a `list`) and sum up all its values. Twice.
  * Avoiding allocating loads of temporary `string`s usually benefits FreeBSD. Here, not so much.
* For the next 2 items, losing the region sorting in favor of an ordered `map` doesn't do much. Additionally, we'll pay a penalty (just how much, I've not looked at yet) when *Computing stats per traveler* for doing this.
  `TMArray`, I've not looked into for this yet. That's a larger-scope project for another time.
* I have pretty low expectations for the last 2 items. Low priority.

Rather than go straight down the rabbit hole to the point of hobbling machines with slow RAM, I'll probably take a break from this for a bit, to come back later.
In the meantime, one place I see potential for improvement is [Reading waypoints for all routes.](https://forum.travelmapping.net/index.php?topic=5349.msg30141#msg30141) There's lots of [unnecessary/redundant `string` construction that can be avoided](https://github.com/TravelMapping/DataProcessing/issues/466).
Beyond that, other than storing objects in contiguous memory (using a `TMArray` as necessary) and whatever advantages that may bring, it seems that most attempts at improving speed are approaching diminishing returns.

Issue #518, opened 2022-04-17
poor graph generation performance on FreeBSD


yakra:
**Near-term**
* Add a commandline switch for single-threaded subgraphs, to retain the benefit from other threaded tasks while staying @ peak efficiency @ 1 SubgraphThread.
* Or better yet, have another go at the [numthr](https://github.com/yakra/DataProcessing/tree/25f3a81b728677ab3a2fd2e04c08fbf2c8d5be44) branch, adding the ability to customize the sweet spot for each task.
      (See also #519)
* **Update:** These 2 bullet points were intended as workarounds for coping with poor performance.
  Addressing #585 will fix that, so no real need for these.
  The `numthr` branch may be worth looking into eventually, but it's nothing specific to graph generation. It can be its own issue if need be.

**Long-term**
* [x] How much does the HGVertex bit field help?
  Not much. Down 29.2 -> 24.7 s @ 1 thread. Right back to its old tricks @ 2 threads, taking 49.3 s.
* [x] (af9d017) Sequentially stored TravelerLists?
* Converting `HGVertex::unique_name` from a `string*` to a `const char*` will save a little indirection.
  Going directly to the `const char*` can save us from unnecessarily loading the `string` into cache from main memory, only to grab its `data()`.
  **Update:** Tracking this bullet point at yakra#270 instead.
* Comment things out & see how much things speed up. Do the same for lab3 and/or lab4 for comparison.
  * [x] traveled graphs
  * [x] I forget whether I actually did comment out **edges** at any point, but now having found the solution, this is no longer important. :)
  * [x] .sql file
* Identify structures likely to be accessed by different threads concurrently. Keep cache lines in mind.
  * [x] `HGEdge::written` is an array of `numthreads` bytes. If there are 64 threads, they could all be on the same cache line. Even though this doesn't appear to be a problem on Linux, but still.
    A bit field _may_ help here.
    * For array indices, do another pass thru the data after `HighwayGraph` construction. An `unsigned int` fits in currently unused space. Init to -1.
    * As an added bonus, it saves RAM:
      * Don't waste 5 bytes for every 3 used.
      * No need for a pointer in each `HGEdge`.
  * [x] `TravelerList::in_subgraph`
  * `HGVertex::s_vertex_num`, `c_vertex_num` & `t_vertex_num` are scattered all over the heap. One big-ol' `unsigned int[vertices.size()][3]` for each thread may help with locality & cache misses.
[...]

Issue #151, opened 2018-12-14
recording per-traveler clinched segments


yakra:
https://github.com/TravelMapping/DataProcessing/blob/b8af79317d9ccdb21896b04da34c0a5e12b09658/siteupdate/python-teresco/siteupdate.py#L1002-L1004
Leaving a reminder to continue to pay attention to this as I pore through the code.

Working from a somewhat outdated version of the HighwayData repo here locally, with [4281202 traveler-segments](http://forum.travelmapping.net/index.php?topic=94.msg12428#msg12428). Just storing the memory addresses of the objects in a container, at 8 bytes a pop, is **> 32** MB. If we were able to store this once instead of twice, that'd be dandy.

From a RAM use perspective considering the overhead of the containers themselves, it makes more sense to store the segments inside a `TravelerList` than to store the travelers inside a `HighwaySegment`.
The former case would require ~231 containers for ~231 travelers; the latter ~723010 containers for ~723010 segments. That's a difference of 722779 containers. With a C++ list taking up 24 bytes on my system (I don't know how much space the containers use in Python), that's **> 16** MB.

Another RAM advantage comes into play if we want to benefit from multithreading in C++. Processing multiple `TravelerList`s concurrently, we need to put a mutex *(a Lock, in Pythonese)* around `HighwaySegment::add_clinched_by`. For any significant time savings, we need to have a unique mutex for each `HighwaySegment` object. Writing only to`TravelerList`s, we don't need these. 723010 total segments * 40 bytes per mutex =	**> 27** MB savings.
Getting rid of the mutex means one process doesn't having to wait for another to write to a given object, means time savings. *Small* time savings in practice, but more than zero.

On the downside is CPU time. Thinking out loud...
When a traveler-segment is matched during .list processing, we must check to see if it's already stored as clinched for that traveler.
• Storing segments inside a `TravelerList` means iterating through longer lists. Starting with a list of 0 elements & ending with a list of, on average, [18533 segments](http://forum.travelmapping.net/index.php?topic=94.msg12428#msg12428) = iterating through a list of 9267 elements on average.
• Storing travelers inside a `HighwaySegment` means iterating through shorter lists, max [128 elements](http://forum.travelmapping.net/index.php?topic=94.msg12428#msg12428).

How much time difference would this work out to in practice? Don't know.
The old RAM vs. CPU trade-off...

I'll have to see what's practicable, and what's needed where in the code...

A couple arguments for using the more RAM-intensive method:
[...]

PR #579, opened 2023-02-11
TravelerList::clinched_segments -> vector, etc.

yakra:
Minor bits & bobs:
* 1cc2b9570cdfbe6a8cd501c3149712c074de8f4b is what it says on the tin.
* 780a65bb456f4755e719dd127308de9f9dcc042c fixes an omission from the original Python implementation of full custom graphs, where only the vertex and edge counts were printed, without the filename. Originally a bit too straightforward a translation from the C++ version apparently; forgot to take into account that this should happen here, and isn't handled by `write_subgraphs_tmg`.

---

616425bc2fb5df01778f43c2d06ccbe15ae94816 closes yakra#230. C++ only.
Saves a bit of RAM in `Route` objects and speeds up *Reading systems list*.
Same idea as in #468, just much lower impact due to far fewer routes than waypoints, fewer AltRouteNames than AltLabels, and the task being so fast to begin with. That and we're not getting rid of any memory copies. All that considered, not too bad though:

machine | OS | before | after | time<br>ratio | % speed<br>increase
-- | -- | -- | -- | -- | --
BT | Ubuntu 16.04 | 0.4318 | 0.3543 | 0.821 | 21.87
lab5 | Ubuntu 18.04 | 0.2058 | 0.1753 | 0.852 | 17.40
lab4 | Ubuntu 18.04 | 0.3694 | 0.3141 | 0.850 | 17.61
lab1 | CentOS 7 | 0.3087 | 0.2846 | 0.922 | 8.47
lab2 | CentOS 7 | 0.3875 | 0.3487 | 0.900 | 11.13
lab3 | CentOS 7 | 0.4946 | 0.4404 | 0.890 | 12.31
bsdlab | FreeBSD 13.0 | 0.6941 | 0.6939 | 1.000 | 0.03

---

And now, saving the best for last, as mentioned in https://github.com/TravelMapping/DataProcessing/issues/151#issuecomment-1413899998...
a90d66477911bdc60f5cedd62f8c315450b9ee34 converts `TravelerList::clinched_segments` from an unordered_set to a vector, that we only add to if `HighwaySegment::add_clinched_by` returns True.
* RAM savings per `TravelerList` of 16 B on FreeBSD, 32 B on Ubuntu 16.04 (DK about Ubuntu 18.04 or CentOS) plus whatever the "hidden" costs of `std::unordered_set` allocated on the heap are.
* Speeds up 3 separate tasks:

### Processing traveler list files
List
We lose a bunch of set insertions in `Route::store_traveled_segments`, instead conditionally appending to a vector.

**Python:**
machine | before | after
| -- | -- | -- |
lab1 | 10.71 | 10.36 | 
lab5 | 8.99 | 8.58 | 
lab2 | 13.01 | 12.44

Underwhelming results in Python. I blame its type system.

### Augmenting travelers for detected concurrent segments
Conc
Two components to the speedup here:
* Iterating through a vector will be less expensive than iterating through an unordered_set. Python benefits from this too. This particular speedup is here to stay. But OTOH...
* Inserting the `TravelerList::clinched_segments` into a vector instead of an unordered_set, for later use when *Computing stats per traveler*. Multi-threaded C++ only. SiteupdateST & Python don't have the extra insertions, because I never changed the way *Computing stats* iterated, because single threaded.
This component will eventually become moot when other branches of the code catch up. Fixing [the region.php rankings bug](https://forum.travelmapping.net/index.php?topic=5379) means changing the way *Computing stats* iterates, means no longer needing to insert the additional elements into `clinched_segments`. So that'll be removed completely, making this task even faster still.

**Python:**
machine | before | after
| -- | -- | -- |
lab1 | 9.81 | 8.46
lab5 | 8.36 | 7.04
lab2 | 11.44 | 10.32

A little more substantial improvement this time, but still not approaching the gains for C++.

### Computing stats per traveler
cst
Multi-threaded C++ only. This is a good example of the benefits of iterating through a vector rather than an unordered_set. [Not much else is going on](https://github.com/TravelMapping/DataProcessing/blob/76589583c4da970b867655e8ca70bb73999201b4/siteupdate/cplusplus/classes/HighwaySegment/compute_stats_t.cpp) in `CompStatsTThread` other than some hash table lookups and arithmetic. The code itself is unchanged for this task; just the container we're iterating through.
This improvement will also be short-lived though. With the upcoming region.php rankings bugfix, *Computing stats per traveler* will cease to exist as its own discrete task.
There are a few different roads I can take here.
* Before I hit the pause button on the bugfix Feb 1, I had 9 different build alternatives to evaluate, and was about to introduce 3 more. Some significantly slower, some about the same speed, some a little faster.
* Meanwhile, working on #151 again has led me to [dust off an old idea and develop it further](https://github.com/yakra/DataProcessing/issues/221#issuecomment-1416854847).
**A new `TMBitset` class would save some RAM and speed up:**
• Processing traveler list files (Route::store_traveled_segments)
• Augmenting travelers for detected concurrent segments
• userlogs (Route::clinched_by_traveler)
• graph generation (edge compression)
• graph generation (traveled graph traveler lists and `clinchedby_code`s)
**The trade-off is slower:**
• computing stats
• writing the `segments` SQL table
...due to increased costs of iterating through this new class.
  * There are a couple potential optimizations I've yet to try out that could lessen the impact on iteration. Where it would ultimately stand relative to iterating through an `unordered_set` remains to be seen.
  * For writing the SQL tables, we can work around this if desired by iterating through `TravelerList::clinched_segments` to write the `clinched` table.
  (Unless we *really* wanna save about 64 MB RAM by clearing these vectors once computing stats (or augmenting concurrencies, after the rankings bugfix is implemented) is finished.)
  We could do this now, in fact.
    * With the conversion of `clinched_segments` to a vector, that's about as efficient as iteration gets! Way faster than an unordered_set or a TMBitset with whatever optimizations it eventually includes.
    * A side benefit of this is not having to create the big `std::vector<std::string> clinched_list` while writing the `segments` table, and all the allocations, re-allocations, and moves that entails. It would play a lot nicer with memory bandwidth. A Good Thing considering the .sql file is written concurrent with graph file generation, which is pretty bandwidth-hungry. Especially in [FreeBSD](https://github.com/TravelMapping/DataProcessing/issues/518), which usually suffers sooner & more greatly from RAM bottlenecks.

### DIFF
--- a/siteupdate/cplusplus/classes/Route/Route.cpp
+++ b/siteupdate/cplusplus/classes/Route/Route.cpp
@@ -219,8 +219,8 @@ void Route::store_traveled_segments(TravelerList* t, std::ofstream& log, std::st
 {	// store clinched segments with traveler and traveler with segments
 	for (unsigned int pos = beg; pos < end; pos++)
 	{	HighwaySegment *hs = segment_list[pos];
-		hs->add_clinched_by(t);
-		t->clinched_segments.insert(hs);
+		if (hs->add_clinched_by(t))
+		  t->clinched_segments.push_back(hs);
 	}
 	if (last_update && t->updated_routes.insert(this).second && update.size() && last_update[0] >= update)
 		log << "Route updated " << last_update[0] << ": " << readable_name() << '\n';
--- a/siteupdate/cplusplus/classes/Route/Route.h
+++ b/siteupdate/cplusplus/classes/Route/Route.h
@@ -6,7 +6,6 @@ class Region;
 class TravelerList;
 class Waypoint;
 class WaypointQuadtree;
-#include <deque>
 #include <mutex>
 #include <unordered_map>
 #include <unordered_set>
@@ -59,7 +58,7 @@ class Route
 	std::string abbrev;
 	std::string city;
 	std::string root;
-	std::deque<std::string> alt_route_names;
+	std::vector<std::string> alt_route_names;
 	ConnectedRoute *con_route;
 
 	std::vector<Waypoint*> point_list;
--- a/siteupdate/cplusplus/classes/TravelerList/TravelerList.h
+++ b/siteupdate/cplusplus/classes/TravelerList/TravelerList.h
@@ -31,7 +31,7 @@ class TravelerList
     start_region start_route start_point end_region end_route end_point
     */
 	public:
-	std::unordered_set<HighwaySegment*> clinched_segments;
+	std::vector<HighwaySegment*> clinched_segments;
 	std::string traveler_name;
 	std::unordered_map<Region*, double> active_preview_mileage_by_region;				// total mileage per region, active+preview only
 	std::unordered_map<Region*, double> active_only_mileage_by_region;				// total mileage per region, active only
--- a/siteupdate/cplusplus/siteupdate.cpp
+++ b/siteupdate/cplusplus/siteupdate.cpp
@@ -1,10 +1,10 @@
 // Tab Width = 8
 
-// Travel Mapping Project, Jim Teresco and Eric Bryant, 2015-2022
+// Travel Mapping Project, Jim Teresco and Eric Bryant, 2015-2023
 /* Code to read .csv and .wpt files and prepare for
 adding to the Travel Mapping Project database.
 
-(c) 2015-2022, Jim Teresco and Eric Bryant
+(c) 2015-2023, Jim Teresco and Eric Bryant
 Original Python version by Jim Teresco, with contributions from Eric Bryant and the TravelMapping team
 C++ translation by Eric Bryant
 
--- a/siteupdate/cplusplus/threads/ConcAugThread.cpp
+++ b/siteupdate/cplusplus/threads/ConcAugThread.cpp
@@ -20,6 +20,6 @@ void ConcAugThread(unsigned int id, std::mutex* tl_mtx, std::list<std::string>* 
 		      {	augment_list->push_back("Concurrency augment for traveler " + t->traveler_name + ": [" + hs->str() + "] based on [" + s->str() + ']');
 			to_add.push_back(hs);
 		      }
-		t->clinched_segments.insert(to_add.begin(), to_add.end());
+		t->clinched_segments.insert(t->clinched_segments.end(), to_add.begin(), to_add.end());
 	}
 }
--- a/siteupdate/python-teresco/siteupdate.py
+++ b/siteupdate/python-teresco/siteupdate.py
@@ -1,9 +1,9 @@
 #!/usr/bin/env python3
-# Travel Mapping Project, Jim Teresco, 2015-2022
+# Travel Mapping Project, Jim Teresco, 2015-2023
 """Python code to read .csv and .wpt files and prepare for
 adding to the Travel Mapping Project database.
 
-(c) 2015-2022, Jim Teresco, Eric Bryant, and Travel Mapping Project contributors
+(c) 2015-2023, Jim Teresco, Eric Bryant, and Travel Mapping Project contributors
 
 This module defines classes to represent the contents of a
 .csv file that lists the highways within a system, and a
--- a/siteupdate/python-teresco/siteupdate.py
+++ b/siteupdate/python-teresco/siteupdate.py
@@ -1173,8 +1173,8 @@ class Route:
         # store clinched segments with traveler and traveler with segments
         for pos in range(beg, end):
             hs = self.segment_list[pos]
-            hs.add_clinched_by(t)
-            t.clinched_segments.add(hs)
+            if hs.add_clinched_by(t):
+                t.clinched_segments.append(hs)
         if self.last_update and self not in t.updated_routes:
             t.updated_routes.add(self)
             if t.update and self.last_update[0] >= t.update:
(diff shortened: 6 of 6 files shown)


yakra:
a90d66477911bdc60f5cedd62f8c315450b9ee34 saves about 270 MiB on Ubuntu 16.04.

yakra:
Another added bonus:
concurrencies.log produced by different runs of siteupdateST are now DIFFable!

michihdeu:
So, only 0.03% speed increase on Jim's machine?

> bsdlab 	FreeBSD 13.0 	0.6941 	0.6939 	1.000 	0.03

And 16 Byte x ~360 = ~5,760 Byte RAM saving?

yakra:
> So, only 0.03% speed increase on Jim's machine?

For 616425bc2fb5df01778f43c2d06ccbe15ae94816, 0.03% speed increase on my FreeBSD machine, same hardware as lab3 & lab4.
You'll see in the same table there's a bit of variation between different machines running the same OS. So noreaster may not be exactly 0.03%, but in the ballpark.
OK, so now I'm curious. Starting a test on nroeaster. :rofl: 1.61% faster! (Table updated.)
Of course, even BiggaTomato, the machine with the most wall time difference, saves a whopping 0.08s. But *it was the right thing to do!*

> And 16 Byte x ~360 = ~5,760 Byte RAM saving?

Yes and no.

**Yes** in that 16 B \* 363 = 5808 B, allocated on the stack, known at compile time. This figure comes from `sizeof(std::unordered_set<HighwaySegment*> - sizeof(std::vector<HighwaySegment*>)` = [40 - 24](https://github.com/yakra/DataProcessing/issues/168) = 16 B.

**No** in that both an [unordered_set](https://cplusplus.com/reference/unordered_set/unordered_set/) and a [vector](https://cplusplus.com/reference/vector/vector/), like many data structures, will have to dynamically allocate some amount of memory, unknown at compile time, on the heap.
* For a simple example, let's look at [Region::allregions](https://github.com/TravelMapping/DataProcessing/blob/cc9da66106a7581974938040418b845dcedaf4b2/siteupdate/cplusplus/classes/Region/Region.h#L51). That's a vector; it's `sizeof` is 24 B.
  * But it has to dynamically allocate an array of `Region` pointers. At 8 B * 510 (including a [dummy](https://github.com/TravelMapping/DataProcessing/blob/1cc2b9570cdfbe6a8cd501c3149712c074de8f4b/siteupdate/cplusplus/siteupdate.cpp#L173-L174)), that's 4080 B.
  * But that's not the whole story. In order to avoid having to reallocate *every time* an item is added, when a vector is resized, it allocates extra memory to allow itself room to grow. In all implementations I've looked at, it doubles in size every time. So its [capacity](https://cplusplus.com/reference/vector/vector/capacity/) is 512 elements.
  * ...making the final size 512*8+24 = 4120 B.
* For a more complex example, consider the `WaypointQuadtree` -- sizeof([all_waypoints](https://github.com/TravelMapping/DataProcessing/blob/1cc2b9570cdfbe6a8cd501c3149712c074de8f4b/siteupdate/cplusplus/siteupdate.cpp#L211-L213)) = 152 B. Not much in and of itself.
  * As `Waypoint*`s are inserted and the quadtree is refined, tens of thousands more `WaypointQuadtree` objects are created. 52,821 of them as of the latest [siteupdate.log](https://travelmapping.net/logs/siteupdate.log), for a total of 8,028,792 B so far.
  * And then, terminal nodes will have a [list](https://cplusplus.com/reference/list/list/) of `Waypoint` pointers. Lists take up more space than vectors; each node contains, in addition to the stored object itself, pointers to the previous & next nodes. 3 pointers * 8 B = 24 B, * 1,238,553 waypoints = 29,725,272 B. Add in those 8 MB for storing the quadtree nodes themselves, and that makes the *real* size of `all_waypoints` 37,754,064 B. **36 MiB!**
* An `unordered_set` will operate somewhere between these 2 extremes. I don't know how exactly they work under the hood (which can differ in different compilers) and don't want to go to the bother of finding out. But they have to manage different [buckets](https://cplusplus.com/reference/unordered_set/unordered_set/bucket/) and arrays of stuff, however that all works; suffice it to say it's more complex than a `vector`. More memory allocated on the heap.
  * Hence mentioning the "hidden cost". Looking at my System Monitor during runs of the old and new executables [on BiggaTomato, it saves about 270 MiB](https://github.com/TravelMapping/DataProcessing/pull/579#issuecomment-1426933955).
  * FreeBSD is a bit different; I can't get as precise a handle on how much RAM is used for each run in [htop](https://htop.dev/). But it does appear to be about the same ballpark.

Other than saving about 270 MiB, the big win here is the speedup.

> btw, ìs there a benefit on still updating Python? Except of having fun and learning....

I'm running with the "continue to update Python in parallel until we decide to officially sunset it" philosophy.
It helps for intelligibility purposes to keep the two versions similar in terms of data structures, functions & algorithms where practicable. Of course, there are enough other big differences throughout the program that, how important *is* this, really?
While still updating Python, what's important is only to keep doing the same datachecks & output the same data -- logs, csvs, .sql file -- the same way.
OTOH, the changes here were so easily done that, hey, why not! Plus I'm always curious to see what the time savings are. That's where the having fun and learning comes in, I guess!

Issue #574, opened 2023-01-26
Optimize data check output


michihdeu:
- [Remove "skip lines"](https://forum.travelmapping.net/index.php?topic=5359.msg30338#msg30338)
`./datacheck.sh: SKIPPING nmpbyregion (no ../nmpfilter/nmpbyregion executable)`
`./datacheck.sh: SKIPPING file copies and DB update`
- [Highlight success message by blank lines](https://forum.travelmapping.net/index.php?topic=5359.msg30334#msg30334)
- [Nix or massivly reduce progress dots](https://forum.travelmapping.net/index.php?topic=5359.msg30350#msg30350)

yakra:
> [Nix or massivly reduce progress dots](https://forum.travelmapping.net/index.php?topic=5359.msg30350#msg30350)

I wrote on the forum:
> I've thought about nixing all the dots for each Route when *Reading waypoints for all routes.*
It'd be nice to not have to whack Page Down 11 times to get past that stuff when siteupdate.log is loaded into `less`.

In a similar vein...
* This would make the output for *Reading waypoints for all routes* look a lot like that of *Reading systems list*.
WRT *Reading systems list*, on the one hand, using progress dots instead of the full `systemname` would slim it down and take less space onscreen.
On second thought though, I'd like to keep this as-is. If something causes a crash at this stage of the program, all I have to do is run `siteupdateST` or `siteupdate -t 1`, and the last system listed is where I need to start looking for problems.
* *Writing near-miss point merged wpt files* has a similar appearance, and is a candidate for slimming down to progress dots. These routines are pretty stable, haven't had any substantial changes in a long time, and probably won't for the foreseeable future, so no need to worry about any specific system causing a crash.
Low priority though:
  * `nmp_merged` files aren't done in `siteupdate --errorcheck` (aka `datacheck`).
  * @jteresco [has](https://github.com/TravelMapping/DataProcessing/issues/565) [proposed](https://forum.travelmapping.net/index.php?topic=5363) removing them from production site updates.
* [ ] Slim down *Processing traveler list files* a bit by only printing the `traveler_name` and omitting the `.list` extension.

michihdeu:
Talking about reading systems.... Do we need the "ignored systems" output?

It's a long list right now..... Ok, Texas is your bad and France my bad but.... do we need it?

@denishanson the five commented jpnh entries are superfluous now, aren't they?

`Ignored comment in systems.csv: #Preview Systems
[...]

PR #584, opened 2023-03-28
Region::compute_stats / output cleanup

yakra:
Closes yakra#189.
Closes yakra#210.
Closes yakra#226.

d7f64b9a199c5c3839dc0409f202e4ead311d750 fixes [the region.php rankings bug](https://forum.travelmapping.net/index.php?topic=5379), and reunifies `CompStatsRThread` and `CompStatsTThread` into a single `CompStatsThread` that iterates by Region. Along the way, bits of related & unrelated major & minor reorganization & cleanup.
Included, `TravelerList` construction only prints the `traveler_name` instead of the whole filename with `.list` extension, saving a little real estate onscreen & in siteupdate.log (https://github.com/TravelMapping/DataProcessing/issues/574#issuecomment-1408924231).
Slightly smaller memory footprint: Each HighwaySegment is 8 B smaller. Each HighwaySystem & Route has one less mutex, saving 40 B on Linux & 8 B on FreeBSD. Each Region is 24 B larger. 

e27917f35980e32a0f50aa42cf949da0105a00d7 removes the [float conversion kluge-around](https://github.com/TravelMapping/DataProcessing/commit/31ccd30c985c95960a3b5b57850d89e7f463ecd8#diff-8a323ef0bce4aba54cfd7eef56fdfeda18216c23c3f5135c13b00a89e5449386) discussed [on the forum](https://forum.travelmapping.net/index.php?topic=5379.msg30521#msg30521).
Remembering having written this kluge in the first place was what immediately tipped me off to the cause of the bug @mapcat reported. It's not actually the cause, but just a band-aid over one of the symptoms. In any case, time for it to go.
In *theory*, this speeds things up a tiny bit in the process, but in practice, not enough to notice. Even in the Python version, the process of *Creating per-traveler stats log entries and augmenting data structure* went from 26.90 s to 26.89 s.
Each TravelerList is 16 B smaller.

The next 5 commits clean up output for #574. Everything within the siteupdate program itself is taken care of in the first 4 of these commits. For siteupdate.sh (AKA datacheck.sh & /fast/tm/datacheck), cae38f10302e0fc32bec300b539892053ff723b1 removes the output when *creating .time files*, but other items still need to be sorted out, so #574 remains open.

89f079c63b88058be822d5f3b35889ddfed7ef21 deserves some special attention. I was pretty surprised by the speedup that removing the progress dots for read_wpt yielded. Two (seemingly) simple lines of code changed. All we stopped doing is inserting 1 character (`.`) into the std output stream per Route & 2 chars (`!\n`) per HighwaySystem, and flushing the output stream in each case. But for all their apparent simplicity, these are some pretty heavyweight operations. Check out the time differences in the chart below. FreeBSD in particular is more than twice as fast at 11, 12 & 19+ threads!

RWpt
Not depicted: lab5 -- its hard disk failed a few days ago, and I'm not satisfied with the write performance of its replacement, which is bad enough to affect benchmarks when writing siteupdate.log with all those dots in it. :(

What makes this even more remarkable is the sheer amount of stuff done during this thread, for writing to stdout to still remain such a large percentage of.
For an incomplete list, there's erasing the filename from `all_wpt_files` (within a mutex), allocating memory & reading the .wpt file into it, Splitting the data into lines & then processing them, removing whitespace, constructing the Waypoint object (split line into fields, create AltLabels & label, parse & validate URL), adding it to the route, inserting it into the quadtree (including looking for colocated points, DUPLICATE_COORDS datacheck, and refining the quadtree as needed), creating HighwaySegments, and performing 27 of the 37 datachecks.


jteresco:
Wow, thanks for all of this.  I'll merge, pull down to noreaster, and we'll use this tonight.  I'll go back to the multithreaded version unless I hear otherwise.

jteresco:
Thanks - recompiled that one too.

PR #587, opened 2023-04-01
HIDDEN_TERMINUS with VISIBLE_DISTANCE fixes

### LINKED ISSUE #583: HIDDEN_TERMINUS with VISIBLE_DISTANCE
Hidden beginning with VISIBLE_DISTANCE:
* **C++** crashes when trying to dereference a null pointer.
  * [x] Move hidden beginning detection into an `else` block after [this](https://github.com/TravelMapping/DataProcessing/blob/3c15c69fc0670e2450117527c2edd54bc68d2956/siteupdate/cplusplus/classes/Route/read_wpt.cpp#L86-L91), then set `last_visible = point_list[0];`.
* **Python:**
  ```py
  [299.1] Performing data checks..Traceback (most recent call last):
    File "./siteupdate.py", line 4180, in <module>
      labels.append(last_visible.label)
  AttributeError: 'NoneType' object has no attribute 'label'
  ```
  * [x] Similarly, set `last_visible = r.point_list[0]` when declared. Should be able to nix the initial `last_visible = None` above.

Hidden end with VISIBLE_DISTANCE:
Both languages fail to detect the error.
* [x] When hidden final point is detected, call `visible_distance` one more time.

**Related:** #360, #582

yakra:
Closes #583.

### DIFF
--- a/siteupdate/cplusplus/classes/Route/read_wpt.cpp
+++ b/siteupdate/cplusplus/classes/Route/read_wpt.cpp
@@ -12,7 +12,7 @@
 void Route::read_wpt(WaypointQuadtree *all_waypoints, ErrorList *el, bool usa_flag)
 {	/* read data into the Route's waypoint list from a .wpt file */
 	std::string filename = Args::highwaydatapath + "/hwy_data" + "/" + rg_str + "/" + system->systemname + "/" + root + ".wpt";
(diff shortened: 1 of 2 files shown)


jteresco:
Thanks.  I'll launch an extra site update this morning wth this.

Issue #591, opened 2023-06-04
Makefile dependencies


yakra:
Dependencies for some targets can change without being reflected in their dependency files. Without becoming known to GNU make. Wacky antics ensue.
Problem is, an object's dependency file is only remade when its primary .cpp file changes. Not when there are changes to its dependencies, or *their* dependencies, which can happen without any changes to the .cpp file.

Let's unpack this a bit, and assume that GNU make is run after each commit.
I've been finishing off the implementation of the `TMBitset` class, and making a cleaner commit history by doing some rebasing, reordering & squashing commits, and adding some finishing touches along the way. None of this is on GitHub yet.

Early on in this process, some minor efficiency improvements were made to [compute_stats.cpp](https://github.com/TravelMapping/DataProcessing/blob/80f27243bf26579c53c6c7a215a5aeb22347317a/siteupdate/cplusplus/classes/Region/compute_stats.cpp).
There are no further changes to this file going forward.
Thus compute_stats.cpp is newer than compute_stats.d, compute_stats.d is remade, compute_stats.o is recompiled, siteupdate & siteupdateST are relinked. So far so good.

8 commits later, TMBitset is introduced.
HighwaySegment.h [changes](https://github.com/TravelMapping/DataProcessing/blob/80f27243bf26579c53c6c7a215a5aeb22347317a/siteupdate/cplusplus/classes/HighwaySegment/HighwaySegment.h#L19) to [include](https://github.com/TravelMapping/DataProcessing/blob/80f27243bf26579c53c6c7a215a5aeb22347317a/siteupdate/cplusplus/classes/HighwaySegment/HighwaySegment.h#L7) TMBitset.cpp. No other files directly include it.
Everything that depends on TMBitset.cpp includes it via HighwaySegment.h. That file was changed, so everything gets recompiled that needs to be. So far so good... *Right?*

The next commit changes both TMBitset.cpp & HighwaySegment.h. Everything gets recompiled once again.
After that, things get a bit more iffy.

The next commit changes only TMBitset.cpp.
The `|=` operator keeps the same return type and arguments, and processes the same data laid out the same way as before; it just does what it does a bit differently. Everything compiles, links & runs fine.
[...]

Issue #589, opened 2023-04-28
nix HGEdge::segment_name


yakra:
During `HGEdge` construction, each edge is assigned a [std::string segment_name](https://github.com/TravelMapping/DataProcessing/blob/80f27243bf26579c53c6c7a215a5aeb22347317a/siteupdate/cplusplus/classes/HighwaySegment/HighwaySegment.cpp#L32-L46) based on all its concurrent active/preview routes. Despite what the comment says, it's not actually used for graph edge labels -- that functionality instead falls to [void HighwaySegment::write_label](https://github.com/TravelMapping/DataProcessing/blob/80f27243bf26579c53c6c7a215a5aeb22347317a/siteupdate/cplusplus/classes/HighwaySegment/HighwaySegment.cpp#L91-L108), which accounts for system restrictions in system/multisystem/fullcustom graph sets. (For example, edge labels in cantch-system.tmg will only list TCHFoo, not AB16, NB2, PE1, NS104, etc.)

Where is this datum actually used?
* [Some commented-out debug text](https://github.com/TravelMapping/DataProcessing/blob/80f27243bf26579c53c6c7a215a5aeb22347317a/siteupdate/cplusplus/classes/GraphGeneration/HGEdge.cpp#L63-L68) and [an unused function (for more debug text)](https://github.com/TravelMapping/DataProcessing/blob/80f27243bf26579c53c6c7a215a5aeb22347317a/siteupdate/cplusplus/classes/GraphGeneration/HGEdge.cpp#L184-L196) that it calls.
  Here, `segment_name` can be replaced with a call to the underlying `segment->segment_name()`, and life goes on.
* That leaves us with [this one active sanity check](https://github.com/TravelMapping/DataProcessing/blob/80f27243bf26579c53c6c7a215a5aeb22347317a/siteupdate/cplusplus/classes/GraphGeneration/HGEdge.cpp#L56-L61), which is [only triggered by HIDDEN_TERMINUS errors.](https://github.com/TravelMapping/DataProcessing/issues/91#issuecomment-1526812024)

We could remove this check (and thus `HGEdge::segment_name` altogether) and nip these cases in the bud before we proceed to the `HGEdge` construction stage. There are a couple different ways to do this...
[...]

PR #592, opened 2023-07-09
TMArray, TMBitset, etc.

### LINKED ISSUE #525: handling nonexistent .list files
Not a problem in production, since by default siteupdate processes .list files found in the UserData repo.
Only a problem if the `-U` or `--userlist` argument is used to specify an invalid username.
Should be an easy enough fix.

* **Python:**
  ```
  [89.9] Processing traveler list files:
  DAVE.list Traceback (most recent call last):
    File "./siteupdate.py", line 3229, in <module>
      traveler_lists.append(TravelerList(t,update,el,args.userlistfilepath))
    File "./siteupdate.py", line 1415, in __init__
      with open(path+"/"+travelername,"rt", encoding='UTF-8') as file:
  FileNotFoundError: [Errno 2] No such file or directory: '/home/yakra/ytm/UserData/list_files/DAVE.list'
  ```
* **C++** can run to completion, reading garbage from a random memory location, zeroing out another random byte in memory in the process.
  ```
  Log file created at: Mon Apr 25 18:49:10 2022
  Incorrect format line (4 or 6 fields expected, found 1): 
  Processed 0 good lines marking 0 segments traveled.
  Clinched Highway Statistics
  Overall in active systems: 0.00 of 953328.25 mi (0.00%)
  Overall in active+preview systems: 0.00 of 1345492.71 mi (0.00%)
  Overall by region: (each line reports active only then active+preview)
  
  Traveled 0 of 257 (0.0%), Clinched 0 of 257 (0.0%) active systems
  Traveled 0 of 87 (0.0%), Clinched 0 of 87 (0.0%) preview systems
  
  Most recent updates for listed routes:
  ```

### LINKED ISSUE #588: canonical segments for graph edges
If a devel system has concurrent segments with an active/preview system and is listed before the a/p system in systems.csv, graph generation breaks.
* Graph edges are never created for segments with devel systems at the front of their concurrency lists.
* Some hidden points thus go from 2 incident edges to 0, and are thus included in collapsed & traveled graphs, with no edges connecting them.

This has never happened in real life AFAIK, but still.
Basing the canonical `HGEdge` segment not on the front of the concurrency list but instead on the first active/preview segment will prevent this from happening.
A quick little function in the `HighwaySegment` class and minor modification to initial `HGEdge` creation could be our insurance policy, hopefully not too expensive.

yakra:
I believe this is all ready to rock. But if you feel like waiting a few days before merging it all in, no harm there.

This does not address the [Makefile issues](https://github.com/TravelMapping/DataProcessing/issues/591). I want to give that some more thought.
You can `make clean`, or if you don't want to recompile *everything*, `rm threads/threads.d classes/Route/read_wpt.d` (from the `cplusplus/` dir) should be sufficient.

What *is* included?

**Lower impact items:**
* Closes #588
* siteupdate.log timestamp for "Making list of travelers". I've occasionally seen this task take a bit of time after a lot of disk activity or cold boot.
* Closes #525
* Misc. minor cleanup

**Big-ticket items:**
* Implements new `TMArray` & `TMBitset` template classes. More on these below.
* Closes yakra#245.

It's possible this closes some older more obscure issues too (probably in yakra/DataProcessing rather than TravelMapping/DataProcessing). If so I'll close them manually as I come across them.

### DIFF
--- a/siteupdate/cplusplus/classes/GraphGeneration/HGEdge.cpp
+++ b/siteupdate/cplusplus/classes/GraphGeneration/HGEdge.cpp
@@ -158,10 +158,10 @@ void HGEdge::collapsed_tmg_line(std::ofstream& file, char* fstr, unsigned int th
 }
 
 // write line to tmg traveled edge file
-void HGEdge::traveled_tmg_line(std::ofstream& file, char* fstr, unsigned int threadnum, std::list<HighwaySystem*> *systems, std::list<TravelerList*> *traveler_lists, char* code)
+void HGEdge::traveled_tmg_line(std::ofstream& file, char* fstr, unsigned int threadnum, std::list<HighwaySystem*> *systems, bool trav, char* code)
 {	file << vertex1->t_vertex_num[threadnum] << ' ' << vertex2->t_vertex_num[threadnum] << ' ';
 	segment->write_label(file, systems);
-	file << ' ' << segment->clinchedby_code(traveler_lists, code, threadnum);
+	file << ' ' << (trav ? segment->clinchedby_code(code, threadnum) : "0");
 	for (HGVertex *intermediate : intermediate_points)
 	{	sprintf(fstr, " %.15g %.15g", intermediate->lat, intermediate->lng);
 		file << fstr;
--- a/siteupdate/cplusplus/classes/GraphGeneration/HGEdge.h
+++ b/siteupdate/cplusplus/classes/GraphGeneration/HGEdge.h
@@ -1,7 +1,6 @@
 class HGVertex;
 class HighwaySegment;
 class HighwaySystem;
-class TravelerList;
 #include <iostream>
 #include <list>
 
@@ -28,7 +27,7 @@ class HGEdge
 	void detach();
 	void detach(unsigned char);
 	void collapsed_tmg_line(std::ofstream&, char*, unsigned int, std::list<HighwaySystem*>*);
-	void traveled_tmg_line (std::ofstream&, char*, unsigned int, std::list<HighwaySystem*>*, std::list<TravelerList*>*, char*);
+	void traveled_tmg_line (std::ofstream&, char*, unsigned int, std::list<HighwaySystem*>*, bool, char*);
 	std::string debug_tmg_line(std::list<HighwaySystem*> *, unsigned int);
 	std::string str();
 	std::string intermediate_point_string();
--- a/siteupdate/cplusplus/classes/GraphGeneration/HGEdge.h
+++ b/siteupdate/cplusplus/classes/GraphGeneration/HGEdge.h
@@ -18,9 +18,9 @@ class HGEdge
 	unsigned char format;
 
 	// constants for more human-readable format masks
-	static const unsigned char simple = 1;
-	static const unsigned char collapsed = 2;
-	static const unsigned char traveled = 4;
+	static constexpr unsigned char simple = 1;
+	static constexpr unsigned char collapsed = 2;
+	static constexpr unsigned char traveled = 4;
 
 	HGEdge(HighwaySegment *);
 	HGEdge(HGVertex *, unsigned char);
--- a/siteupdate/cplusplus/classes/GraphGeneration/HGVertex.h
+++ b/siteupdate/cplusplus/classes/GraphGeneration/HGVertex.h
@@ -4,7 +4,6 @@ class Region;
 class Waypoint;
 #include <list>
 #include <string>
-#include <unordered_set>
 
 class HGVertex
 {   /* This class encapsulates information needed for a highway graph
--- a/siteupdate/cplusplus/classes/GraphGeneration/HighwayGraph.cpp
+++ b/siteupdate/cplusplus/classes/GraphGeneration/HighwayGraph.cpp
@@ -109,11 +109,11 @@ HighwayGraph::HighwayGraph(WaypointQuadtree &all_waypoints, ElapsedTime &et)
 			{	new HGEdge(w->vertex, HGEdge::collapsed);
 				continue;
 			}
-			else if ((w->vertex->incident_c_edges.front() == w->vertex->incident_t_edges.front()
+			if   (	 (w->vertex->incident_c_edges.front() == w->vertex->incident_t_edges.front()
 			       && w->vertex->incident_c_edges.back()  == w->vertex->incident_t_edges.back())
 			      || (w->vertex->incident_c_edges.front() == w->vertex->incident_t_edges.back()
-			       && w->vertex->incident_c_edges.back()  == w->vertex->incident_t_edges.front()))
-				new HGEdge(w->vertex, HGEdge::collapsed | HGEdge::traveled);
+			       && w->vertex->incident_c_edges.back()  == w->vertex->incident_t_edges.front())
+			     )	new HGEdge(w->vertex, HGEdge::collapsed | HGEdge::traveled);
 			else {	new HGEdge(w->vertex, HGEdge::collapsed);
 				new HGEdge(w->vertex, HGEdge::traveled);
 				// Final collapsed edges are deleted by HGEdge::detach via ~HGVertex via HighwayGraph::clear.
--- a/siteupdate/cplusplus/classes/GraphGeneration/HighwayGraph.cpp
+++ b/siteupdate/cplusplus/classes/GraphGeneration/HighwayGraph.cpp
@@ -91,22 +91,14 @@ HighwayGraph::HighwayGraph(WaypointQuadtree &all_waypoints, ElapsedTime &et)
 				w->vertex->visibility = 2;
 				continue;
 			}
-			// if edge clinched_by sets mismatch, set visibility to 1
-			// (visible in traveled graph; hidden in collapsed graph)
-			// first, the easy check, for whether set sizes mismatch
-			if (w->vertex->incident_t_edges.front()->segment->clinched_by.size()
-			 != w->vertex->incident_t_edges.back()->segment->clinched_by.size())
-				w->vertex->visibility = 1;
-			// next, compare clinched_by sets; look for any element in the 1st not in the 2nd
-			else for (TravelerList *t : w->vertex->incident_t_edges.front()->segment->clinched_by)
-				if (!w->vertex->incident_t_edges.back()->segment->clinched_by.count(t))
-				{	w->vertex->visibility = 1;
-					break;
-				}
 			// construct from vertex this time
 			--ce; --cv;
-			if (w->vertex->visibility == 1)
-			{	new HGEdge(w->vertex, HGEdge::collapsed);
+			// if edge clinched_by sets mismatch, set visibility to 1
+			// (visible in traveled graph; hidden in collapsed graph)
+			if (w->vertex->incident_t_edges.front()->segment->clinched_by
+			 != w->vertex->incident_t_edges.back()->segment->clinched_by)
+			{	w->vertex->visibility = 1;
+				new HGEdge(w->vertex, HGEdge::collapsed);
 				continue;
 			}
 			if   (	 (w->vertex->incident_c_edges.front() == w->vertex->incident_t_edges.front()
--- a/siteupdate/cplusplus/classes/GraphGeneration/HighwayGraph.cpp
+++ b/siteupdate/cplusplus/classes/GraphGeneration/HighwayGraph.cpp
@@ -62,7 +62,7 @@ HighwayGraph::HighwayGraph(WaypointQuadtree &all_waypoints, ElapsedTime &et)
 		counter++;
 		for (Route *r : h->route_list)
 		  for (HighwaySegment *s : r->segment_list)
-		    if (!s->concurrent || s == s->concurrent->front())
+		    if (s == s->canonical_edge_segment())
 		    { ++se; 
 		      new HGEdge(s);
 		      // deleted by HGEdge::detach via ~HGVertex via HighwayGraph::clear
--- a/siteupdate/cplusplus/classes/GraphGeneration/HighwayGraph.cpp
+++ b/siteupdate/cplusplus/classes/GraphGeneration/HighwayGraph.cpp
@@ -298,7 +298,7 @@ void HighwayGraph::write_master_graphs_tmg()
 	travelfile << "TMG 2.0 traveled\n";
 	simplefile << vertices.size() << ' ' << se << '\n';
 	collapfile << cv << ' ' << ce << '\n';
-	travelfile << tv << ' ' << te << ' ' << TravelerList::allusers.size() << '\n';
+	travelfile << tv << ' ' << te << ' ' << TravelerList::allusers.size << '\n';
 
 	// write vertices
 	unsigned int sv = 0;
@@ -314,7 +314,7 @@ void HighwayGraph::write_master_graphs_tmg()
 		}
 	}
 	// now edges, only write if not already written
-	size_t nibbles = ceil(double(TravelerList::allusers.size())/4);
+	size_t nibbles = ceil(double(TravelerList::allusers.size)/4);
 	char* cbycode = new char[nibbles+1];
 			// deleted after writing edges
 	cbycode[nibbles] = 0;
@@ -329,7 +329,7 @@ void HighwayGraph::write_master_graphs_tmg()
 			  if (!(e->written[0] &  HGEdge::traveled))
 			  {	e->written[0] |= HGEdge::traveled;
 				for (char*n=cbycode; n<cbycode+nibbles; ++n) *n = '0';
-				e->traveled_tmg_line(travelfile, fstr, 0, 0, &TravelerList::allusers, cbycode);
+				e->traveled_tmg_line(travelfile, fstr, 0, 0, TravelerList::allusers.size, cbycode);
 			  }
 	    default:	for (HGEdge *e : v.incident_s_edges)
 			  if (!(e->written[0] &  HGEdge::simple))
@@ -342,8 +342,8 @@ void HighwayGraph::write_master_graphs_tmg()
 	  }
 	delete[] cbycode;
 	// traveler names
-	for (TravelerList *t : TravelerList::allusers)
-		travelfile << t->traveler_name << ' ';
+	for (TravelerList& t : TravelerList::allusers)
+		travelfile << t.traveler_name << ' ';
 	travelfile << '\n';
 	simplefile.close();
 	collapfile.close();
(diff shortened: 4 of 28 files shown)


yakra:
## TMArray

This is a simple wrapper around a C array, with an allocation function, destructor, and iteration.
Right now, it's only used for storing `TravelerList` objects, but eventually will be used for darn near everything.

**Background:**
In the very early days of writing the program, I had issues with invalid pointers.
For example, a `Waypoint` object is created in the `Route::point_list` [vector](https://cplusplus.com/reference/vector/vector/), NMPs are found (this point points to others & others point to this), and a pointer is inserted into the quadtree. Later on, after adding more items into `point_list`, it runs out of capacity & has to reallocate, and all the waypoints are moved, invalidating all the pointers.
Sure, we know how many lines are in a .wpt file, and can thus call [std::vector::reserve](https://cplusplus.com/reference/vector/vector/reserve/), obviating reallocation. If I was even aware of this function at that stage of the game, this solution didn't occur to me.
But wait, there's more!
As `HighwaySystem` objects are created in a master system list vector, `Route` objects are created, containing a pointer to their parent system. Resizing the vector as more systems are created means, again, invalidating all those pointers. This is probably the first place I ran into this problem, at the very start of getting siteupdate.cpp off the ground.
Here though, even calling `vector::reserve` will not work; `HighwaySystem` objects contain a [mutex](https://cplusplus.com/reference/mutex/mutex/). Mutexes cannot be moved, therefore objects containing them cannot be moved either. Vectors require that their contents be moveable in order for reallocation to work, so attempting to create a `vector<HighwaySystem>` fails to compile.
I ended up keeping a kinda "Pythonic" approach, creating `new` objects on the heap individually, and saving pointers to them in the various containers. The objects stay where they're created, and no need to worry about points going invalid.
I used this approach for consistency's sake with most containers, probably even ones where there was no risk of invalidating pointers and no mutex preventing storage in a vector.

**Motivation:**
It's time to move away from this. Storing objects in contiguous memory is more efficient in several ways.
* More predictable memory access patterns; consistently skip ahead `sizeof(foo)` bytes when iterating.
* Cache locality. More cache hits.
* Stop wasting RAM storing a bunch of unnecessary pointers.
* It opens the door to more efficient algorithms that take advantage of pointer arithmetic. For example...
  * The TMBitset class. More on this below.
  * Replacing [these two clunky functions](https://github.com/TravelMapping/DataProcessing/blob/80f27243bf26579c53c6c7a215a5aeb22347317a/siteupdate/cplusplus/classes/HighwaySystem/HighwaySystem.cpp#L138-L150).
  * Concurrency detection could benefit from this, not that it's all that slow right now.
  * Converting the three per-traveler [regional tables](https://github.com/TravelMapping/DataProcessing/blob/bc04537138af75f558bd9b4481e1fbe324c5e04a/siteupdate/cplusplus/classes/TravelerList/TravelerList.h#L36-L38) to ordered maps could speed up userlogs and/or stats CSVs, at the cost of a potential minor slowdown when computing stats.
  * ...and much more.

If `vector` won't work, raw C arrays or, more conveniently, a new custom container class will. We can keep using range for loops via the begin() and end() functions.

Secondly, the destructor cleans up allocated memory when a `TMArray` object goes out of scope. This helps make siteupdate "valgrind clean", stopping valgrind from reporting a bunch of memory leaks, without having to manually worry about manually `delete`ing everything. LOL, and it doesn't contribute to the `Total run time` figure if it happens at the very end after that, as we exit `main`. If that's arguably cheating, it's a kind I'm OK with. :)

**Does it use less RAM?**
* The object itself is 16 B, compared to 24 B for a vector or [list](https://cplusplus.com/reference/list/list/).
* For the memory it dynamically allocates to store its data, it depends on how we use it.
  * A `TMArray<Foo>` is no less efficient than a `vector<Foo>`. A `vector<Foo>` can be *as* efficient if we first `reserve` space before populating it, but most of the time that's not done; usually items are just added and the vector reallocates, resulting in unused extra space.
  * We are able to get rid of unnecessary pointers in our existing `vector<Foo*>` containers though.
  * A `list<Foo>` would contain pointers to the prev/next items at each list node.

**Is it faster?**
No faster than a `vector<Foo>`, but faster than what we have now.
* Reduced indirection
* As noted above, RAM & cache benefits
* Indirectly, the potential for newer better algorithms.

yakra:
## TMBitset

This is a replacement for [unordered_set](https://cplusplus.com/reference/unordered_set/unordered_set/)s of pointers.
It stores one bit for each `TravelerList` in the system, with on/off indicating whether or not it's in the set, rather than 64 bits for each `TravelerList*` in the set. Info for up to 512 travelers can fit in a single cache line, compared to only 8 64-bit pointers.
I went to kind of ridiculous lengths to Micro-Optimize All The Things.

**Is it faster?**
Yes.
* Insertion:
  No more taking time hashing, mucking about with buckets & checking for duplicate values.
  The distance of a `TravelerList`'s bit from the start of the bit array = the distance of its memory address from `TravelerList::allusers.data`. A little arithmetic, and we can flip one bit, and BOOM!
  Speeds up:
  * Processing traveler list files
  * Augmenting travelers for detected concurrent segments
* Checking membership:
  No more taking time hashing & mucking about with buckets.
  Via the simple arithmetic above, we find a bit and see if it's 0 or 1.
  Speeds up:
  * Creating per-traveler stats logs and augmenting data structure
* Comparing contents of one set to another:
  Before, unless the two sets were [different sizes](https://github.com/TravelMapping/DataProcessing/blob/bc04537138af75f558bd9b4481e1fbe324c5e04a/siteupdate/cplusplus/classes/GraphGeneration/HighwayGraph.cpp#L96-L99), we had to [iterate through the contents of one](https://github.com/TravelMapping/DataProcessing/blob/bc04537138af75f558bd9b4481e1fbe324c5e04a/siteupdate/cplusplus/classes/GraphGeneration/HighwayGraph.cpp#L101) (slow!) and [check each item's membership in the other](https://github.com/TravelMapping/DataProcessing/blob/bc04537138af75f558bd9b4481e1fbe324c5e04a/siteupdate/cplusplus/classes/GraphGeneration/HighwayGraph.cpp#L102) set (slow!) [until](https://github.com/TravelMapping/DataProcessing/blob/bc04537138af75f558bd9b4481e1fbe324c5e04a/siteupdate/cplusplus/classes/GraphGeneration/HighwayGraph.cpp#L104) we find one missing.
  Now, we have two arrays of equal length, with each TravelerList represented in the same position in both -- `memcmp` those babies and BOOM!
  Speeds up:
  * Compressing collapsed edges
* Unions of sets:
  [Before](https://github.com/TravelMapping/DataProcessing/blob/bc04537138af75f558bd9b4481e1fbe324c5e04a/siteupdate/cplusplus/classes/GraphGeneration/HighwayGraph.cpp#L270-L274): iterate thru members of each set, check membership, add to a new master list, set membership to `true`.
  Now: Create a new master TMBitset, and instantly add up to 64 items at once via a Bitwise Or.
  Speeds up:
  * Traveled graph generation
* Iteration:
  I was worried this might be slower than iterating an `unordered_set`. Turns out it's faster. Win!
  Speeds up:
  * Computing stats
  * Traveled graph generation
  * `segments` SQL table

**Does it use less RAM?**
* The object itself is 32 B, compared to an `unordered_set`, which is 40 B on FreeBSD, 48 B on my CentOS boxes, and 56 B on Ubuntu. At 1,149,116 segments, we save 8.8, 17.5, and 26.3 MB respectively. Not bad. :)
* For the memory it dynamically allocates to store its data, the short answer is **yes**, but it's more complex; it depends on the underlying data. With relatively few items in each set, an `unordered_set` can be more efficient. With relatively few total items to choose from, a `TMBitset` can be more efficient, which is what we have. An `unordered_set<Foo*>` uses 64 bits for every item in the set, whereas a `TMBitset` allocates 1 bit for every item whether in the set or not.
  With 377 travelers in the system, a `TMBitset` allocates 48 bytes. I *could* have had devel segments avoid allocating this space, but that'd only save 901 KB, so I opted to keep it simple. Thus all 1,149,116 segments allocate these 48 B, for a total of 52.6 MB. With 8,393,470 entries in the `clinched` DB table times 8 B, that's 64 MB.
  Or, on a per-segment basis *(including devel segments because 1, it's the more apples-to-apples comparison because I allocate this memory with TMBitset and 2, it's the more conservative figure more favorable to unordered_set)* 58 B vice 48.
  Plus, all of this doesn't even take into account all the extra RAM `unordered_set` has to allocate behind the scenes to get stuff to work under the hood.
* Overall, 80f2724 -> 634da8b, System Monitor tells me I'm saving about 325 MB.

**There are [2 template parameters](https://github.com/TravelMapping/DataProcessing/blob/634da8ba67e30a71e13e78cbb9edeca34f37ba73/siteupdate/cplusplus/templates/TMBitset.cpp#L8). What's that all about?**
[The 1st one](https://github.com/TravelMapping/DataProcessing/blob/634da8ba67e30a71e13e78cbb9edeca34f37ba73/siteupdate/cplusplus/classes/HighwaySegment/HighwaySegment.h#L19), `item` is pretty straightforward. It's the data type included (or not) in the set.
The 2nd one, `unit`, controls the size of the data chunks that are allocated and processed, and allows us to tine-tune how things behave.
* Smaller units like `uint8_t` or `unsigned char` minimize the amount of unused RAM allocated.
  Right now with 377 travelers, it makes no difference; we're using 48 B either way.
* Larger units like `uint64_t` or `unsigned long` allow the `|=` (bitwise or) operator to work more efficiently.
* Unit size also affects how iteration behaves under the hood. The sweet spot depends on the underlying data; small units could be fastest, maybe large ones, maybe somewhere in between. I chose `uint32_t` for the current application.

**Do I have bigger plans for this class?**
Not so much. I only see 2 further applications right now. Neither uses `unordered_set`, so I'd expect to see less benefit than we see for graph generation with this pull request.
* Vertex membership for subgraphs already uses an ad hoc bit field. There could be a tiny speedup in zeroing out all these bits between graphs, but not much else. The primary gain here is cleaner code.
* HGEdge membership for subgraphs uses an array of `numthreads` bools, similar to what TravelerLists did until now. The potential benefits are in memory locality & cache hits. There's a bit of a heavier lift to get this one into place. Low priority, though I'd still like to try it eventually.

Even if there's not much more to be done, at least this has already provided a good speed-up in a number of areas.

yakra:
The first 2 tasks benefit from faster set insertions.

List
* lab3 does a better job of scaling up to 23 threads now.
* lab4 already made it to 23, but now scales all the way to 24, outperforming lab2 despite its lower clock speed, memory bandwidth & cache.
* bsdlab starts out, as before, comparable to lab3, but starts to suffer from FreeBSD's memory bandwidth issues & doesn't scale up to more threads as well. There's still some improvement though, and this effect isn't as pronounced as with other tasks. I've some ideas to speed this up which may or may not pan out; low priority.

Conc
* Good performance with FreeBSD here, outperforming lab3 up to \~6 threads, comparable above that.

yakra:
Iteration at work.

Stat
I was pleasantly surprised to find iterating `HighwaySegment::clinched_by` faster with a `TMBitset` than an `unordered_set`.

yakra:
Checking set membership.

User
* Big improvements here. LOTS of lookups. For every route in a system a user has traveled, we iterate though its segments, check whether each is `clinched_by` a traveler, and if so, add its mileage to the traveler's per-route total.
* For FreeBSD, this helps at the extreme low end, but at 4+ threads we start to saturate the memory bus, and do about the same as before.
  This is very similar to its behavior with graph generation, where it's most efficient at just 1 thread, wall time increases quickly then plateaus from 4-10 threads, ramps up again till 18 threads, then falls off, ending up a bit slower than the first plateau.
* I've some long-term low-priority ideas that could potentialy help out here too.

yakra:
Comparing contents of two sets:
The `!=` operator at work.

Comp
* FreeBSD does great here; compare its performance to lab3 (CentOS 7) & lab4 (Ubuntu 18.04) on the same hardware.
* Big improvement for lab4; it's not terribly far behind bsdlab now.

yakra:
Traveled graph generation benefits from iteration and unions of sets.

Graph
As seems to be the pattern with [FreeBSD, graph generation](https://github.com/TravelMapping/DataProcessing/issues/518) improvements appear to save a tiny bit of time at 1 thread, and just not help out above that.
Still though, I remain cautiously optimistic. There are a number of avenues to explore targeting memory locality, `HGEdge` [alignment](https://github.com/TravelMapping/DataProcessing/issues/589) & [size](https://github.com/TravelMapping/DataProcessing/issues/589#issuecomment-1540084836), cache hits, etc.

Graph_Linux
Same thing, zoomed in to just the Linux machines. Performance seems to max out at ~6 threads. (For now at least?)

yakra:
Total
Overall, siteupdate's peak efficiency on bsdlab is at just 1 thread. :( The good news is, it's faster @ 1 thread now than it was @ 2 before.
Graph generation is a big piece of the pie, and successfully speeding it up should make the picture look much better.
I didn't think of this till just now, but an interesting chart to look at would be total wall time in `--errorcheck` mode, with only userlogs being a real poor perfomer. It'll take me another day to produce the data for that though.

jteresco:
Wow, this looks great.  Thanks for all the effort on this.

rickmastfan67:
So, @yakra, this is why you've been ignoring your TX threads on the forums, eh? LOL. ;)

yakra:
Thanks! I'm particularly proud of [TMBitset](https://github.com/TravelMapping/DataProcessing/blob/68513556f661e832c47e7369f277943105846ed5/siteupdate/cplusplus/templates/TMBitset.cpp). :grinning:
Leveled up a bit at C++ in the process. This pull request includes the first uses of `constexpr`, `friend`, `calloc`, `aligned_alloc`, and [placement new](https://en.cppreference.com/w/cpp/language/new#Placement_new). Partial template specialization was a headache, but now I know some stuff. Tried out inheritance before settling on the [TMBImpl template class](https://github.com/TravelMapping/DataProcessing/blob/68513556f661e832c47e7369f277943105846ed5/siteupdate/cplusplus/templates/TMBitset.cpp#L79-L87). Experimented with [thread_local](https://en.cppreference.com/w/cpp/language/storage_duration) variables, but they didn't make the final cut.

> I didn't think of this till just now, but an interesting chart to look at would be total wall time in `--errorcheck` mode, with only userlogs being a real poor perfomer. **It'll take me another day to produce the data for that though.**

Or not. I've just started the benchmark script on bsdlab, the last machine left. That would take about 14 hours *if* graph generation were enabled, slowing overall speed down to a crawl. Without it? I'll guess "5 more hours" & see how close that is. May not be awake by then.

Having ticked off a bunch of boxes at #518, I just need to "sweep up the shop floor". Gotta delete a couple dozen local branches that were rebased, reordered, squashed, cherrypicked, or went nowhere. Some of the code that didn't make the final cut, I'd still like to save for future reference. Probably paste it into yakra#245.
After that, I plan to return to the forum shortly.

HA! Speaking of which, I see @rickmastfan67 [wrote](https://github.com/TravelMapping/DataProcessing/pull/592#issuecomment-1627882040):
> So, @yakra, this is why you've been ignoring your TX threads on the forums, eh? LOL. ;)

Yes, exactly that. LOL. *All* the threads.
I wanted laser-like focus on this sub-project, and chose to tune out other distractions while working on it. That, and the day job leaves a lot less time & energy for the project than I've had in the past. I think it's been what, a couple-few months since I've logged in / viewed the forum?
I've had a few forum threads open in other browser tabs this entire time, waiting for me to get back to them. If they're reloaded every time I restart firefox, that may give a false positive & show me as active on the forum, but yeah no, I've been MIA.
Will be back soon, I know I've *at least* one unread PM. And I know @mapcat has driven the newly-opened realignment of NS104 west of Antigonish.

rickmastfan67:
> I've had a few forum threads open in other browser tabs this entire time, waiting for me to get back to them. If they're reloaded every time I restart firefox, that may give a false positive & show me as active on the forum, but yeah no, I've been MIA.

Yep, every time you restarted Firefox, it pinged the forums and reloaded said threads.  Thus, since you were logged in, updated your profile and said you were logged in.

Let's just say, you've possibly missed 20+ threads for your areas. :p lol. ;)

And the threads you had pinned in firefox, probably will not properly show what posts (if any) are 'new' to them, as after the first refresh of it due to the browser restart, that 'new' button will go bye-bye.

So, in the future, might I suggest creating a toolbar folder and save bookmarks to threads in that?  That way, you'll not loose the 'new' part, as well as will not ping the forums saying you're active in you leave the threads pinned in Firefox as 'active'.  Then, once you've addressed the thread, just delete the bookmark. :)

rickmastfan67:
Oh, and the "Show unread posts since last visit." link will 99.9999% be useless to you as well, due to those pinned threads you left open in Firefox sadly.  You'll have to use the "[Click here to try all unread topics](https://forum.travelmapping.net/index.php?action=unread;all;start=0)." link instead and dig thru it to find threads in your areas, as that will be the only way to properly see what's unread to your account.

yakra:
> And the threads you had pinned in firefox, probably will not properly show what posts (if any) are 'new' to them, as after the first refresh of it due to the browser restart, that 'new' button will go bye-bye.

No big deal on these probably. Most of them weren't very active. I can take the `.msg65535#new`, change it to  `.msg65535#msg65535` go by what's onscreen and look at the post dates if necessary.

> Oh, and the "Show unread posts since last visit." link will 99.9999% be useless to you as well, due to those pinned threads you left open in Firefox sadly.

For whatever reason, that link's never really been on my radar. I'll probably just stick to my usual MO...
Middle-click each sub-board to open in a new tab. IIUC, the "new" notifications here won't be accurate due to my "logins", so just click everything. There's a small number of sub-boards thankfully.
Look for topics with unread posts -- these should be tagged `new` regardless of my logins. And then middle-click a bunch more tabs. :P

> You'll have to use the "[Click here to try all unread topics](https://forum.travelmapping.net/index.php?action=unread;all;start=0)." link instead

Interesting; don't think I was aware of this one. Is it linked from the main page?

michihdeu:
> > So, @yakra, this is why you've been ignoring your TX threads on the forums, eh? LOL. ;)
> 
> Yes, exactly that. LOL.

Good to know 😃 I was a little bit worried meanwhile... You didn't miss much but lost your pole position in the "Top 10 Posters" ranking 🤣

yakra:
https://github.com/TravelMapping/DataProcessing/pull/592#issuecomment-1627784490
> **Do I have bigger plans for this class?**
Not so much. I only see 2 further applications right now. Neither uses `unordered_set`, so I'd expect to see less benefit than we see for graph generation with this pull request.
> * [Vertex membership](https://github.com/yakra/DataProcessing/issues/251) for subgraphs already uses an ad hoc bit field. There could be a tiny speedup in zeroing out all these bits between graphs, **but not much else**. The primary gain here is cleaner code.

[On further thought...](https://github.com/yakra/DataProcessing/issues/251)
* Creating vertex sets for multi-region (and country & continent) and multi-system graphs could benefit from `|=`ing together multiple region or system sets.
* Full custom graphs could create intersections of these multi-region & multi-system sets with a new `&` or `&=` operator.

---

@michihdeu wrote:
> Good to know 😃 I was a little bit worried meanwhile... You didn't miss much but lost your pole position in the "Top 10 Posters" ranking 🤣

I take that to mean, not that I've dropped all the way out of the top 10, but just lost the # 1 spot. :grin: Honestly, I've kinda wanted that to happen for a while.
And unless things have changed drasticaly since I was last around... congrats! :grinning:

Issue #595, opened 2023-07-20
Add option for alternate name of hwy_data directory


jteresco:
As part of an attempt to get a preliminary railway version of TM up and running, I noticed that the directory name `hwy_data` is hard-coded into the site update program.  This could become a command-line parameter, or possibly we could make `hwy_data` the default name when the repository is called `HighwayData` and `rail_data` when the repository is called `RailywayData`.

jteresco:
An alternative is to rename it as plain `data` in both repositories, with the idea that it's obviously "highway data" or "rail data" based on the name of the repository.  This would eliminate any need for future names like `bike_data` or `ski_data` or `trail_data` if TM continues to expand into other types of travel.

yakra:
Plain `data` sounds reasonable. Keep it simple.

jteresco:
I think I'd like the site update to look for `data` by default.  I don't think it's worth renaming it in the `HighwayData` repository.  Git should be able to maintain history if we do, but I worry about getting everyone who has a clone up to date, especially when they have partially completed work not yet committed to the origin repository.  We should able to add a symbolic link`data` -> `hwy_data`.  The other repositories new data repositories can easily enough just have them renamed as vanilla `data`.

jteresco:
I'm going ahead and making this change since I have a good chance to do it tonight.  Just in C++.

yakra:
OK. Doesn't look like this would cause merge conflicts with anything I have in the pipeline.
Looks like 54fecf1e42a51cae358d12a5bf1f8b2cf4a0e935 gets the 3 crucial files here. I'll take a closer look in a bit.

The `crawl_hwy_data` function still works of course, but may not have the most appropriate name anymore.
`crawl_repo_data`?
`crawl_route_data`?
`crawl_data`?

jteresco:
I was thinking about that as well.  I'm not sure what term works best..

jteresco:
https://forum.travelmapping.net/index.php?topic=5647.msg31783;topicseen#msg31783

yakra:
May be worthwhile to add an option to specify the data subdirectory, for compatibility with HighwayData commits before [fe7ce91](https://github.com/TravelMapping/HighwayData/commit/fe7ce91cbcf06be9f68c296335aa5c91ce66e050).

---

@jteresco [wrote](https://github.com/TravelMapping/DataProcessing/issues/595#issuecomment-1656984724):
> Just in C++.

[...]

PR #607, opened 2023-09-10
TMArray all the things, etc.

### LINKED ISSUE #594: handling invalid data in graph CSVs
https://github.com/TravelMapping/DataProcessing/blob/c51fd20e1f0b896d072bef7ba7740a8e260f8f2b/siteupdate/cplusplus/tasks/subgraphs/fullcustom.cpp#L60-L69

This `continue` breaks out of the [inner loop](https://github.com/TravelMapping/DataProcessing/blob/c51fd20e1f0b896d072bef7ba7740a8e260f8f2b/siteupdate/cplusplus/tasks/subgraphs/fullcustom.cpp#L60), not the [outer one](https://github.com/TravelMapping/DataProcessing/blob/c51fd20e1f0b896d072bef7ba7740a8e260f8f2b/siteupdate/cplusplus/tasks/subgraphs/fullcustom.cpp#L9).
Nasty stuff ensues, like a [double](https://github.com/TravelMapping/DataProcessing/blob/c51fd20e1f0b896d072bef7ba7740a8e260f8f2b/siteupdate/cplusplus/tasks/subgraphs/fullcustom.cpp#L65) [free](https://github.com/TravelMapping/DataProcessing/blob/c51fd20e1f0b896d072bef7ba7740a8e260f8f2b/siteupdate/cplusplus/tasks/subgraphs/fullcustom.cpp#L69), and subgraph generation trying to use an [invalid pointer](https://github.com/TravelMapping/DataProcessing/blob/c51fd20e1f0b896d072bef7ba7740a8e260f8f2b/siteupdate/cplusplus/tasks/subgraphs/fullcustom.cpp#L66).

Break it. Observe it. Fix it.

* [ ] I haven't looked at  the Python version, but it surely has a similar bug.
* Similar checks for invalid data are also appropriate for:
  * [ ] multiregion
  * [ ] multisystem
  * [ ] system

### LINKED ISSUE #602: argument parsing bug
https://github.com/TravelMapping/DataProcessing/blob/1282b3824ab3c457d1ef334234d9232476000f85/siteupdate/cplusplus/classes/Args/Args.cpp#L29
->
```c++
#define ARG(N,S,L) ( n+N < argc && (!strcmp(argv[n],S) || !strcmp(argv[n],L)) )
```

### LINKED ISSUE #606: Data check for double quotes in systems.csv
A `"` in system names [crashes findroute.php](https://github.com/TravelMapping/Web/issues/763). 

@jteresco [wrote](https://github.com/TravelMapping/Web/issues/763#issuecomment-1681420378):
> I'd prefer disallowing double quotes and possibly a hard error if and when that happens.

yakra:
### The *Big Ticket Item* here is converting a bunch of containers to use the `TMArray` class.
Closes yakra#260. As alluded to in https://github.com/TravelMapping/DataProcessing/pull/592/#issuecomment-1627779898:
* Makes siteupdate "valgrind clean" -- valgrind won't report any more memory leaks, even if it was valid in-use data deleted on termination of the program anyway.
* Saves RAM:
  * A `TMArray` is 16 B, vice 24 B for a `std::vector` or 24 or (less commonly) 16 for a `std::list`.
  * Member variables `HighwaySystem::is_valid` and `Region::is_valid` no longer needed; constructors `throw` if in an invalid state instead.
  * Containers store objects themselves, not pointers to objects floating around on the heap.
  * No more empty space proactively allocated by `std::vector`.
* Faster:
  * Storing stuff sequentially = more consistent & predictable memory access patterns.
  * Cache locality. More cache hits.
  * Improved algorithms taking advantage of pointer arithmetic.

### Along the way, some more minor improvements:
* Close #602
* Close #606
* Tick off a few boxes in https://github.com/TravelMapping/DataProcessing/issues/595#issuecomment-1673864517
* Cut down on translation units and #include directives by bundling several string-related functions together into tmstring.cpp
* Refactor lots of stuff out of main siteudate.cpp for brevity & readability purposes. This has been on my to-do list for a while.
  * Region::read_csvs
  * HighwaySystem::systems_csv
  * Most multithreaded tasks
  * Route and label logs
  * routedatastats.log
  * TravelerList::get_ids
  * WaypointQuadtree::nmplogs
* Minor efficiency improvements enabled by pointer arithmetic
  * HighwaySystem::route_index -> Route::index
  * HighwaySystem::con_route_index -> ConnectedRoute::index
  * Concurrency detection optimization
* Close yakra#246
* read_wpt & Waypoint constructor efficiency & organization
  * Simplify loop counters & array indexing when processing lines
  * Close yakra#240
  * Close yakra#256
  * Close yakra#239
* Close #594
* Sort regions.csv lines before creating Region objects
  * Sort by address, not code string
    * userlogs
    * stats csvs
    * routedatastats.log breakdown by system
  * No sorting at all in:
    * allbyregionactiveonly & allbyregionactivepreview
    * routedatastats.log breakdown by region
  * In practice, there's no noticeable difference in execution time here, but still. :)

**Farväl gamle vän**
Python fixes for #594 are included. These were authored before Jim posted [this](https://github.com/TravelMapping/DataProcessing/issues/595#issuecomment-1659458104), which I consider the "official" end of siteupdate.py. I don't plan on making more Python updates in the future, unless I feel like trying something out for purposes of learning the language.


yakra:
RWpt

About half the improvement here comes from the TMArray conversion itself.
Rather than allocate each `new` `Waypoint` and `HighwaySegment` individually, and check if the new vector size exceeds its capacity & re-allocate the vector as needed, we just allocate the `Route::points` and `Route::segments` `TMArray`s once and we're good to go. Placement `new` is called for each location in the array.
The rest of the speedup comes from 3 secondary improvements. In descending order of their impact:
* yakra#256
  Avoid unnecessary std::string construction & allocation for URLs; process in-place in original char buffer.
  With this one in particular, I was hoping to see more improvement for FreeBSD at the high end. Ah well. At least there's some improvement up to 11 threads.
* yakra#240
  The distance update for VISIBLE_DISTANCE & LONG_SEGMENT datachecks uses the stored segment length, rather than recomputing it.
* [Simplify loop counters & array indexing when processing lines](https://github.com/TravelMapping/DataProcessing/blob/d6b94a3213b6226948cfb8d28a827afd7e161c4e/siteupdate/cplusplus/classes/Route/read_wpt.cpp#L53-L62).

You may notice there's a new computer in the mix, [toma](https://chrono.fandom.com/wiki/Toma_Levine). It's a ThinkPad Yoga with an i5-7200U CPU, running Ubuntu 22.04 LTS. RAM I believe is dual-channel DDR4; haven't looked into its clock yet. Big thanks to @flognap for gifting me this trusty steed, in addition to hooking me up with a whole bloody server farm back in 2019.
I assume the lackluster scaling above 2 threads is because the CPU has 2 physical cores, and 4 "logical" hyperthreaded cores. Once we're above the 2 physical cores, there's just not much silicon (ALUs, FPUs, whatever) left to handle the additional workload.
Another factor could be 3 MB last-level cache, vice 8 MB on lab1, lab5 & BiggaTomato.
We'll see this effect again in other benchmarks, while some other tasks scale more gracefully.

yakra:
NmpS
NmpL

For writing nmp_merged .wpt files, there was a small performance bump, but I'm not bothering with a chart as it'll barely even show up.

yakra:
Conc

About 30% of the speedup (on BiggaTomato at least :grin:) comes from the TMArray conversion. We iterate directly through sequential systems, routes, and segments, rather than iterate through a container of pointers, following them to *wherever* on the heap, and *then* going to town.

Beyond that, this routine can benefit from pointer arithmetic.

The obvious quick-n-dirty solution is to do a drop-in replacement of [Route::find_segment_by_waypoints](https://github.com/TravelMapping/DataProcessing/blob/1f92c6cc4499d5f0641ad91c505747f1b5d108b8/siteupdate/cplusplus/classes/Route/Route.cpp#L124-L129):
```c++
HighwaySegment* Route::find_segment_by_waypoints(Waypoint *w1, Waypoint *w2)
{	return	(std::abs(w2 - w1) == 1)
	? (w1<w2 ? w1 : w2) - points.data + segments.data
	: 0;
}
```
There's no more need to iterate & test both endpoints, as we know that the distance of the segment from the beginning of the `segments` array = the distance of its first waypoint from the beginning of the `points` array.

**We can do better than that, though!**
We can eliminate the function call altogether and check for matching waypoints inline, as we know `w2` must either be `w1+1` or `w1-1` in order for there to be a concurrent segment.
The [`if (other)`](https://github.com/TravelMapping/DataProcessing/blob/1f92c6cc4499d5f0641ad91c505747f1b5d108b8/siteupdate/cplusplus/tasks/concurrency_detection.cpp#L16) gets replaced by [`if (w2 == w1+1)`](https://github.com/TravelMapping/DataProcessing/blob/93cc681163252f25c90e2c5debe97841c8167f72/siteupdate/cplusplus/tasks/concurrency_detection.cpp#L15) and [`else if (w2 == w1-1`...](https://github.com/TravelMapping/DataProcessing/blob/93cc681163252f25c90e2c5debe97841c8167f72/siteupdate/cplusplus/tasks/concurrency_detection.cpp#L22)
The conditionals that would have been in `Route::find_segment_by_waypoints` *(whether compiled as jumps or `CMOV`s; I haven't looked into how this shakes out in practice)* are gone, taken care of by the Waypoint pointer passed to a new [HighwaySegment::add_concurrency](https://github.com/TravelMapping/DataProcessing/blob/93cc681163252f25c90e2c5debe97841c8167f72/siteupdate/cplusplus/classes/HighwaySegment/HighwaySegment.cpp#L36-L52) function [carved out of the innermost code block](https://github.com/TravelMapping/DataProcessing/blob/1f92c6cc4499d5f0641ad91c505747f1b5d108b8/siteupdate/cplusplus/tasks/concurrency_detection.cpp#L17-L32).
Then there's [a useless extraneous subexpression](https://github.com/TravelMapping/DataProcessing/commit/bf525b59a6a4b382d435d8ef0c6c796eb672fcad#r123631650); if we imagine that had already been removed, we're left with pretty much the same number of conditional jumps, but reordered to short-circuit earlier.

**And finally, [induction variables](https://en.wikipedia.org/wiki/Induction_variable).**
The 1st point of a route starts the 1st segment, the 2nd point starts the 2nd segment, etc.
Knowing this, now that waypoints are stored sequentially, as we iterate through routes we can [set a pointer to the first waypoint](https://github.com/TravelMapping/DataProcessing/blob/93cc681163252f25c90e2c5debe97841c8167f72/siteupdate/cplusplus/tasks/concurrency_detection.cpp#L9), and then [increment it](https://github.com/TravelMapping/DataProcessing/blob/93cc681163252f25c90e2c5debe97841c8167f72/siteupdate/cplusplus/tasks/concurrency_detection.cpp#L24) in lockstep as we iterate through segments.
No need to keep going into memory and looking up the waypoints from the `HighwaySegment` object.

yakra:
Intg

This is a pretty quick task, so any variations in runtime will be a bigger piece of the pie, making for more jagged graphs. There's also a bit more variation than other tasks that take a comparable amount of time, such as Searching for near-miss points.
That said, a decent amount of improvement is apparent.

yakra:
List

None of the improvements here came from the TMArray conversion. The only TMArray we're iterating through is TravelerList::allusers, which was converted in #592. A TravelerList is a big structure and we're only accessing a couple different widely separated parts of it, so there's not likely to be much of any memory locality / cache hit benefit.

The changes affecting this should all be in 9a177e7b3cedbee9c516aa69fa1cb664f1bf8ebc.
* [This](https://github.com/TravelMapping/DataProcessing/commit/9a177e7b3cedbee9c516aa69fa1cb664f1bf8ebc#diff-dd7b762a45d86c2a97f155349a22b5708ca404d010a4ec6929cb0160aaacaae2R94) may save us a memory copy. **Hm. Could this be it?**
* We [ignore whitespace or "comment" lines](https://github.com/TravelMapping/DataProcessing/commit/9a177e7b3cedbee9c516aa69fa1cb664f1bf8ebc#diff-dd7b762a45d86c2a97f155349a22b5708ca404d010a4ec6929cb0160aaacaae2R108-R112) before [stripping whitespace from the end of the line](https://github.com/TravelMapping/DataProcessing/commit/9a177e7b3cedbee9c516aa69fa1cb664f1bf8ebc#diff-dd7b762a45d86c2a97f155349a22b5708ca404d010a4ec6929cb0160aaacaae2R113-R117) now, saving a few instructions.
* While stripping whitespace from the end, we lose a couple [endchar > lines[l]](https://github.com/TravelMapping/DataProcessing/commit/9a177e7b3cedbee9c516aa69fa1cb664f1bf8ebc#diff-dd7b762a45d86c2a97f155349a22b5708ca404d010a4ec6929cb0160aaacaae2L108-L109) checks (because [this](https://github.com/TravelMapping/DataProcessing/commit/9a177e7b3cedbee9c516aa69fa1cb664f1bf8ebc#diff-dd7b762a45d86c2a97f155349a22b5708ca404d010a4ec6929cb0160aaacaae2R90) happens now), but this is small potatoes. A couple quick instructions, plus the branch would have been predicted correctly pretty much 100% of the time.
* Maybe this new revision is friendlier to L1 instruction cache? Who knows, man.

That said, this is all a pretty small piece of the the pie compared to processing the fields in a line and crediting travelers with clinched segments, which involves a bit of looping, a good chunk of hash table lookups, and tons of branching.
I'm surprised this sped up as much as it did. Look at the delta for BT & Toma especially.

FreeBSD is interesting here. It looks like we're doing worse with the new commit, but are we really?
I re-ran the benchmarks from 2-7 threads in preparation for doing a `git bisect` ([like I did for CompStatsThread on BiggaTomato](https://github.com/TravelMapping/DataProcessing/commit/93cc681163252f25c90e2c5debe97841c8167f72#diff-51440ed5d85c95e3ed1f9ac32e05d980fe6b9a1834516c81a949957aca341edfR39-R48)) and wasn't able to produce the same results; the newer commit was faster about 1/2 the time instead.
This hardware usually has more noise in the benchmark data than the other machines, and this effect is especially pronounced with FreeBSD.
What we see is probably just the result of noise in the data while we're dealing with small differences down near the margin of error.

yakra:
Hash
Logs

A couple single-threaded tasks where iterating systems & routes is a big piece of the pie.
Once again, quite small differences in a rather quick task.
We're very close to margin of error, as we see in the 2nd chart. Toma & bsdlab measured a bit slower. Lab4 is about 1/3 ms faster, its bar shrinking by 1 pixel. ;P

yakra:
Augm

A small but consistent speedup for the Ubuntu machines.

Augm_Ubuntu

Same thing again with the clutter out of the way, Ubuntu only.
What could be happening here? Well, we [iterate through each traveler's `clinched_segments` and access their concurrency lists](https://github.com/TravelMapping/DataProcessing/blob/d6b94a3213b6226948cfb8d28a827afd7e161c4e/siteupdate/cplusplus/threads/ConcAugThread.cpp#L12-L14). Each `clinched_segments` array will have lots of runs of many successive segments added by each .list line. These segments being stored sequentially now makes access to their concurrency lists more efficient.
A secondary benefit is, with many concurrency just being duplexes, we similarly get sequential memory accesses along the concurrent route too.

yakra:
Stat

A bit similar to the last case. While we're not iterating through `HighwaySystem::routes`, we do iterate through `Region::routes`, which is almost as good. With single-region systems being commonplace, each region's `routes` array will have long runs of pointers to successive `Route` objects, with occasional minor disruptions from multi-region routes.
From there, we iterate through sequential `Segment`s in each `Route`.

yakra:
Data

yakra:
User

Iterate systems. Iterate ConnectedRoutes. Iterate ConnectedRoute `roots`, which correlates to `HighwaySystem::routes` storage -- just not 1:1 of course.
The biggest impact is probably when we [iterate through each route's segments](https://github.com/TravelMapping/DataProcessing/blob/d6b94a3213b6226948cfb8d28a827afd7e161c4e/siteupdate/cplusplus/classes/Route/Route.cpp#L153-L158) in order to [find a user's traveled mileage](https://github.com/TravelMapping/DataProcessing/blob/d6b94a3213b6226948cfb8d28a827afd7e161c4e/siteupdate/cplusplus/classes/TravelerList/userlog.cpp#L76-L77).

Like I mentioned in https://github.com/TravelMapping/DataProcessing/pull/592/#issuecomment-1627785857, I've some long-term ideas to improve efficiency and *try to* bring FreeBSD's scaling under control, but they're low priority.
I'm not super optimistic though. From what I've seen so far, there's not much noticeable difference, and even combining all the ideas together may not do much.

yakra:
Edge

Iterate systems, routes, segments; access each segment's `waypoint1`, `waypoint2`, `route` & `concurrent` pointers.
While the amount of time saved looks pretty consistent on Linux, FreeBSD is closer to margin of error and actually comes out looking worse. As mentioned above, FreeBSD has a lot more noise in its time measurements.

yakra:
Graph_Linux

There's an ever-so slight improvement on just the Linux machines.
Zoomed in to just Linux here; [FreeBSD is way off the top of the graph](https://github.com/yakra/DataProcessing/assets/13720877/3b4a7022-4891-4092-b821-93eade4b3106).
A bit unexpected -- there's not much correlation between the order edges are accessed by the time we do so, and the order of the underlying route segments, even if they are closer to one another in memory.

yakra:
Total
Total_linux
ErCh

Overall, a modest speed boost across all machines.
Even for FreeBSD, though it's more apparent in `--errorcheck` mode, as nmp_merged files & graph generation in particular introduce a lot of noise & unpredictability into the data.

yakra:
Finally, a few selected SQL tables. Not all are affected by the TMArray conversion, and those that are aren't affected much.

SQL_crr
**connectedRouteRoots**
* **OT1H,** there's not much going on other than iteration, through systems, connected routes, and chopped routes. Percentage-wise, it's a pretty good speed-up.
* **OTOH,** it's a very quick task, a tiny blip on the radar. Look at that Y axis -- we save only a fraction of a millisecond.

SQL_rtes
The **routes** table contains quite a bit more data per route, and the gap narrows considerably.

SQL_wpts
**Waypoints**

SQL_seg
While writing the **segments** table, we spend some time generating strings for use in the `clinched` table later on, thus the time spent iterating becomes a smaller percentage of the total.

yakra:
Thought I saw a potential bug while reviewing the code changes.
False alarm.
Ready to rock.

rickmastfan67:
@yakra Does this mean you'll now show up on the forum and deal with the 500,000 new posts that have been made dealing with TX & NY? LOL! :P 

Sorry, just had to do this. LOL!

yakra:
HA! Wondered if you might say that. I LOLed.
Yes, I do plan to come back to the forum now.
No more "quick fixes that will only take a couple hours" for the time being. We saw how this one turned out. :P

I do have lots of ideas for speeding up graph generation that I'd love to try out at some point, and hopefully get its FreeBSD performance under control, but thankfully I'm not deluding myself into thinking that's a quick little operation. :sunglasses:

jteresco:
Going for it ahead of tonight's site updates.  Thanks for all of this!

Issue #590, opened 2023-05-14
Site update parameter to use different NMP thresholds


jteresco:
Could add a command-line parameter to the site update programs to allow different NMP thresholds.

https://forum.travelmapping.net/index.php?topic=5314.msg31206#msg31206

It would be unlikely we would change these for production site update runs, but runs could be made by individuals to either use a smaller threshold to limit false positives or a larger threshold to search for unaligned intersection points that are farther away than our default.

yakra:
Even [if siteupdate.py is deprecated](https://github.com/TravelMapping/DataProcessing/issues/595#issuecomment-1659458104) and this option won't make it into the Python version, I think it still makes sense to be consistent and have both a short & long switch for each argument.
* `-N` seems an obvious choice for the short one.
* What about the long one? `--nmp-threshold`? `--nmp-thresh`?

michihdeu:
@yakra Sorry, I didn't realize that your [response on the forum](https://forum.travelmapping.net/index.php?topic=5314.msg32786#msg32786) was to my request. I think that `LOOKS INTENTIONAL` would mean coords being off by exactly 0.000001°. However, I meant coords being of by < 0.01mi.

I manually checked more than 10,000 NMP pairs in France over the past months. I did it region by region and either fixed them or marked FP. I was surprised that there were so many true positives. I always made a second (even more dull) run for each region to catch errors I made during the first run, and there where quite some with very small offset.
So, my French regions are done now - except of FRA-COR being maintained by Si. There are 11,285 FPs. However, I fear that there are some NMPs with very small offset I've marked FP by accident. **Is it possible to make a one-time run with very small NMP threshold for me? All coords being of by ~ 0.01mi?** I would download that file and upload to HDX. The next run can be as usual again. It's fine when it is just tm-master.nmp.

@yakra  or @jteresco is it possible?

jteresco:
I am sure I can modify and compile up a temporary version of siteupdate (or just add the parameter while I'm doing it) and run with whatever threshold you want.  In the runs now, it's set at a tolerance to find points within 0.0005 degrees latitude and longitude.

michihdeu:
thanks @jteresco I think that 0.0001° would be great 😄

jteresco:
@michihdeu I ran with the 0.0001 threshold but not on the production server.  You just want tm-master.nmp?

michihdeu:
[...]

PR #612, opened 2023-12-24
C++ -N flag, #590

jteresco:


yakra:
Just saw this. Off to work in a few mins; can give a closer look late evening.

yakra:
...Or not! This all looks great; thanks for doing it.
No conflicts with anything I have in the pipeline, I believe.
Thankfully, `WaypointQuadtree::near_miss_waypoints` already has a `tolerance` argument, which makes life easier.

yakra:
OH SHIT. Why did I merge this? Can we undo?
Hard reset & force push?

michihdeu:
Should we mention the [parameters](https://github.com/TravelMapping/DataProcessing/blob/master/siteupdate/cplusplus/classes/Args/Args.cpp#L31) in the [readme](https://github.com/TravelMapping/DataProcessing/blob/master/README.md)? Maybe with an example?
Are they also available when running data check? If so, it would be nice to have the hint [here](https://github.com/TravelMapping/DataProcessing/blob/master/RUNNING.md).

yakra:
Would also need the parameters added for site update.sh/datacheck

yakra:
`//printf("Starting NmpSearchThread %02i\n", id); fflush(stdout);`
wasn't commemted back out, so there will be some text on screen and in siteupdate.log that we don't want.

for the sake of a clean commit history, can we do a quick hard reset & force push before others [update their forks](https://github.com/TravelMapping/DataProcessing/forks?include=active&page=1&period=2y&sort_by=last_updated)?

jteresco:
> Can we undo? Hard reset & force push?

Go for it if you'd like.  I can take a look a little later otherwise.

yakra:
Not 100% sure I have the access privileges to do that, via GitKraken at least. GitHub web interface may be another story. I can take a look in about 3 hours.

yakra:
Now that I'm done kicking myself for being too hasty on the merge, I realize that editing & rebasing would give me credit (or "blame", as git calls it) for Jim's code, which is probably the greater evil here. Let's give Jim the cred he deserves for every other line in his commit.
I've made a new commit to master via the web interface.

yakra:
I went ahead and deleted the `datadir` branch, @ 54fecf1e42a51cae358d12a5bf1f8b2cf4a0e935.
@jteresco, both `archivegraphs` & `commonupdate` are merged into master. OK to delete?

jteresco:
Thanks!  And sounds good on deleting the stale branches.

PR #619, opened 2024-03-31
.list processing & DB speedups

### LINKED ISSUE #466: lines[l], orig_line, trim_line

yakra:
Closes yakra#253
Closes #466
Closes yakra#278
Closes yakra#213

### DIFF
--- a/siteupdate/cplusplus/classes/HighwaySegment/HighwaySegment.cpp
+++ b/siteupdate/cplusplus/classes/HighwaySegment/HighwaySegment.cpp
@@ -26,13 +26,6 @@ std::string HighwaySegment::str()
 {	return route->readable_name() + " " + waypoint1->label + " " + waypoint2->label;
 }
 
-bool HighwaySegment::add_clinched_by(size_t t)
-{	clin_mtx.lock();
-	bool result = clinched_by.add_index(t);
-	clin_mtx.unlock();
-	return result;
-}
-
 void HighwaySegment::add_concurrency(std::ofstream& concurrencyfile, Waypoint* w)
 {	HighwaySegment& other = w->route->segments[w - w->route->points.data];
 	if (!concurrent)
--- a/siteupdate/cplusplus/classes/HighwaySegment/HighwaySegment.cpp
+++ b/siteupdate/cplusplus/classes/HighwaySegment/HighwaySegment.cpp
@@ -44,11 +44,6 @@ void HighwaySegment::add_concurrency(std::ofstream& concurrencyfile, Waypoint* w
 	other.concurrent = concurrent;
 }
 
-std::string HighwaySegment::csv_line(unsigned int id)
-{	/* return csv line to insert into a table */
-	return "'" + std::to_string(id) + "','" + std::to_string(waypoint1->point_num) + "','" + std::to_string(waypoint2->point_num) + "','" + route->root + "'";
-}
-
 std::string HighwaySegment::segment_name()
 {	/* compute a segment name based on names of all
 	concurrent routes, used for graph edge labels */
--- a/siteupdate/cplusplus/classes/HighwaySegment/HighwaySegment.h
+++ b/siteupdate/cplusplus/classes/HighwaySegment/HighwaySegment.h
@@ -17,13 +17,11 @@ class HighwaySegment
 	double length;
 	std::list<HighwaySegment*> *concurrent;
 	TMBitset<TravelerList*, uint32_t> clinched_by;
-	std::mutex clin_mtx;
 
 	HighwaySegment(Waypoint*, Route*);
 	~HighwaySegment();
 
 	std::string str();
-	bool add_clinched_by(size_t);
 	void add_concurrency(std::ofstream&, Waypoint*);
 	std::string csv_line(unsigned int);
 	//std::string concurrent_travelers_sanity_check();
--- a/siteupdate/cplusplus/classes/HighwaySegment/HighwaySegment.h
+++ b/siteupdate/cplusplus/classes/HighwaySegment/HighwaySegment.h
@@ -23,7 +23,6 @@ class HighwaySegment
 
 	std::string str();
 	void add_concurrency(std::ofstream&, Waypoint*);
-	std::string csv_line(unsigned int);
 	//std::string concurrent_travelers_sanity_check();
 
 	// graph generation functions
--- a/siteupdate/cplusplus/classes/Route/Route.cpp
+++ b/siteupdate/cplusplus/classes/Route/Route.cpp
@@ -248,16 +248,16 @@ void Route::con_mismatch()
 			       (con_route->banner.size() ? con_route->banner : "(blank)"));
 }
 
-void Route::mark_label_in_use(char* label)
+void Route::mark_label_in_use(std::string& label)
 {	unused_alt_labels.erase(label);
-	labels_in_use.insert(label);
+	labels_in_use.insert(move(label));
 }
 
-void Route::mark_labels_in_use(char* label1, char* label2)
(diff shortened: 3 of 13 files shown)


yakra:
## 44d62f0 ReadListThread mutex efficiency

**Before:**
Every HighwaySegment had a mutex, which was locked when adding a traveler to its `clinched_by` set during .list processing.
Overkill.
With > 1.2 M segments in the system, the odds of 2 being credited for the same traveler concurrently are vanishingly small, even *if* many routes aren't traveled at all & others are used much more frequently.
With ~ 101 K routes in the system, the odds of 2 being credited for the same traveler concurrently are still very small, even if yadda yadda.

**After:**
Locking & unlocking a mutex for every segment is expensive.
Instead, handle this at the route level: Lock `Route::mtx`, call `Route::store_traveled_segments` & credit the traveler for all their segments on that chopped route in the .list line, then unlock `Route::mtx`.
This reduces the number of lock/unlock operations by a factor of 15.9.
`Route::mtx` is already used for locking `labels_in_use` & `unused_alt_labels`, but reusing this rather than adding a new one for locking `HighwaySegment::clinched_by` has little to no effect on contention.
Besides saving RAM, another advantage is we can keep on consolidating, and put the calls to `store_traveled_segments` inside the same lock as `mark_label(s)_in_use`. This last bit doesn't save much time though.

**RAM:**
This lets us scrap the HighwaySegment mutex completely, reducing `sizeof(HighwaySegment)`.
1228371 segments...
\* `sizeof(std::mutex)` of 8 on FreeBSD = 9.4 MB.
\* `sizeof(std::mutex)` of 40 on Linux = 47.9 MB.

List_mtx
Big improvement for FreeBSD! It's competitive with CentOS now, even outperforming it on the low end.

**Secondary benefits:**
Even where the code itself doesn't change, smaller HighwaySegment objects can yield modest performance boosts, I presume due to probability of different data we need being on the same cache line.
This effect is most noticeable on BiggaTomato, my oldest machine with the least memory bandwidth.
A couple selected tasks where the effect is most visible:
* [Augmenting travelers for detected concurrent segments](https://github.com/TravelMapping/DataProcessing/assets/13720877/7aa0f5b1-1baf-4667-8ef5-27132ccba858)
* [Creating per-traveler stats logs and augmenting data structure](https://github.com/TravelMapping/DataProcessing/assets/13720877/6a31e165-6715-490c-9cd2-1c1d29b58abc)

yakra:
## f5a4e85 reduce string construction

**3 problems:**
1. `trim_line` is a copy of the .list line with leading & trailing whitespace trimmed off, for use in userlog messages.
   This is always constructed, but only needed a small percentage of the time.
2. `fields` consists of pointers into the original `char` buffer, with inter-field whitespace zeroed out.
   The label hash tables and `labels_in_use`, `unused_alt_labels` & `duplicate_labels` sets use `std::string`s for keys.
   This means any lookups, additions or deletions involve constructing a `std::string` out of a `char*`, which gets a bit redundant.
3. `orig_line` is a carbon copy of the original .list line, with all whitespace intact, for use in splitregion .list files.
   We *could* use the original `char*` buffer for this, but that's altered & used for `fields` instead.

**3 solutions:**
1. Use `get_trim_line()` to allocate once, as soon as needed. If needed > once, `get_trim_line()` remembers that a string was created previously, and returns it.
2. Create `fields` as `std::string`s from the get-go, avoiding the unnecessary conversions. This leaves the original `char` buffer intact, meaning...
3. The C string from the original buffer can be used for splitregion .list files. No need to copy into `orig_line` anymore.

**Bugfix:**
Along the way, fixed a bug that flagged tabs as invalid characters in userlog error messages for invalid fields.

List_str_zoom
A couple variations of the same chart:
* [Same Y-axis scale as the previous & next posts, for easy side-by-side comparison in separate browser tabs or whatevs.](https://github.com/TravelMapping/DataProcessing/assets/13720877/e943905c-07f8-47cd-b151-4ed13618d6cb)
* [Just the CentOS machines. Older(?) string.h here, without small string optimization.](https://github.com/TravelMapping/DataProcessing/assets/13720877/5e592d45-93de-4225-9709-0978435fc0f4)

yakra:
## Both commits combined, before & after
List_both

jteresco:
Ready to merge, or is more on the way?

yakra:
Ready to merge.

yakra:
I do have a couple more commits on the way now.

yakra:
# DB table speedups

## waypoints (5e8ebb8)

First, let's get rid of `Waypoint::point_num` entirely; it's not needed.
While this doesn't reduce `sizeof(Waypoint)` due to alignment requirements, it does lay the foundation for smaller Waypoints in the future.

The only change to the `waypoints` table routine is we no longer write `point_num` to memory in the Waypoint object.
While this is objectively faster, in practice it's not enough to notice. The effect is obscured by noise in the data.

w
There's no difference between 5e8ebb8 & 0ff0af1. I almost left 0ff0af1 off this chart, but the FreeBSD example especially illustrates what the margin of error can be.

---

## segments

### phase 1 (5e8ebb8)

* With `point_num` no longer stored in the Waypoint object, we can use an induction variable that's kept synced up as we iterate through Routes.
* This allows us to not only avoid looking up `point_num` from a Waypoint, but also looking up `waypoint1` and `waypoint2` from a HighwaySegment. This is a Good Thing; I'm looking at removing those variables in the future.
* Scrap the `HighwaySegment::csv_line` function completely, in favor of inserting its constituent components into `sqlfile` stream.
  * Writing 6 strings & 3 integers is more expensive than writing one string...
  * ...but this is outweighed by not having to construct 11 temporary `std::string` objects.

### phase 2 (0ff0af1)

When constructing `clinched_list` for the `clinched` table afterward, similarly avoid all the `std::string` construction & concatenation, and instead create a `std::pair` out of `segment_num` and `t->traveler_name.data()`.

s

---

## clinched (0ff0af1)

This table takes 1.5-2x as long to write now, but that's the price we pay for greatly speeding up the `segments` table.
The slowdown here + the speedup in the `segments` table hints at the pros and cons of avoiding string construction and doing more `ofstream` insertions, the combined effects of which we saw in segments phase 1.
It's not an exact comparison though. Different numbers of string constructions & file writes, and we weren't creating a vector of pairs there.
* Before: Write 3 strings to `sqlfile`.
  *(It could have easily been one string, with the parentheses & newline included when originally constructed, but hey. \*shrug\*)*
* After: Write 4 strings + 1 integer that has to be converted to string as we do so.

c
Same vertical scale as the clinched table above and the 2 tables combined below, for context.
[Same thing zoomed in.](https://github.com/TravelMapping/DataProcessing/assets/13720877/654ef98f-1445-492c-9dad-5e018a1c85c6) Similar to the waypoints table, 5e8ebb8 had no effect on the clinched table, and we can see the noise in the data in the first 2 commits.

---

## segments + clinched combined time

Total

yakra:
Ready to merge.

jteresco:
will use this for tonight's site update, thanks @yakra as always for all the work on this!

rickmastfan67:
@yakra when you get the chance, can ya check your PM's on the forum.  Thanks. :)

yakra:
Haha, been thinking about how I need to get back there soon.

rickmastfan67:
> Haha, been thinking about how I need to get back there soon.

hehehe.  Some people over there think you've passed along to the next life already. :P (j/k of course)

PR #617, opened 2024-03-20
avoid DISCONNECTED_ROUTE false negatives, etc.

yakra:
9f95b0d is what it says on the tin.
5fb976b closes #610. Siteupdate avoids DISCONNECTED_ROUTE false negatives while going to greater lengths to make sure routes are connected to the best of its ability.
4cad17d is what it says on the tin.

---

Earlier commits were part of #615.
**TLDR:**
* HIDDEN_JUNCTION datacheck takes data from devel systems into account now.
  I wanted to get this working before doing any work on railway datachecks.
* Datachecks out of graph gen = no more graph gen in errorcheck mode. [**Way faster.**](https://github.com/TravelMapping/DataProcessing/pull/615#issuecomment-1971466973)
  Plus fixed some errorchecks that weren't done in errorcheck mode before. Oops.

Closes #433
Closes yakra#242

Issue #585, opened 2023-03-28
C++ siteupdate: Use of sprintf, deprecated on MacOS/XCode


jteresco:
When compiling on MacOS, the C++ site update code generates pages of warnings:

<pre>siteupdate.cpp:527:4: warning: 'sprintf' is deprecated: This function is provided for compatibility reasons only.  Due to security concerns inherent in the design of sprintf(3), it is highly recommended that you use snprintf(3) instead. [-Wdeprecated-declarations]
                        sprintf(fstr, ": %.2f mi", cr->mileage);
                        ^
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/stdio.h:188:1: note: 'sprintf' has been explicitly marked deprecated here
__deprecated_msg("This function is provided for compatibility reasons only.  Due to security concerns inherent in the design of sprintf(3), it is highly recommended that you use snprintf(3) instead.")
</pre>

I'm ok leaving it or replacing our `sprintf` calls with `snprintf` calls.  A little of what I saw when searching around on this is that there are also some more C++-specific ways to achieve the same result.

Compiler version:

<pre>
Apple clang version 14.0.0 (clang-1400.0.29.202)
Target: arm64-apple-darwin21.6.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
</pre>

yakra:
[fprintf](https://cplusplus.com/reference/cstdio/fprintf/) may be able to get rid of most or even all of the `sprintf` calls altogether.

The downside is that it uses C's [FILE](https://cplusplus.com/reference/cstdio/FILE/) rather than C++'s [std::ofstream](https://cplusplus.com/reference/fstream/ofstream/), so it'd involve a little reinvention  of the wheel.

yakra:
## Great news:
sprintf is a performance killer. Yes, this *is* great news.
Getting rid of it inner loops is **the Magic Bullet for fixing poor FreeBSD performance. *Including graph generation.***

Most uses format a string and immediately plop it into a file.
Here, we can use `ofstream.precision`, occasionally `setf` and `unsetf`, and more traditional `<<` operators.
* **graph generation**
  This finally opens the floodgates & allows graph generation to scale to a higher number of threads on FreeBSD.
[...]

PR #622, opened 2024-05-15
TMBitsets for subgraph membership, etc

### LINKED ISSUE #614: devel systems in subgraph CSVs
The `HighwayGraph` is composed of only data in active/preview systems.

Right now, if devel systems are listed in [fullcustom.csv](https://github.com/TravelMapping/HighwayData/blob/master/graphs/fullcustom.csv), [multisystem.csv](https://github.com/TravelMapping/HighwayData/blob/master/graphs/multisystem.csv) or [systemgraphs.csv](systemgraphs.csv), siteupdate doesn't do anything special to detect them.

**What happens?**
I just made `cannss-system-simple.tmg` to find out. Haven't tried the other CSVs, but I expect the same behavior.
I got a perfectly valid file that loaded up fine in the HDX, with all the vertices & edges contained in the other, active, systems in Nova Scotia.

A bit counter to expectation -- one would either expect *all* of cannss be included *(which we can't do because the graph structure is a/p only)*, or only the bits that are a/p *in that system* -- which is to say, nothing.

**Easy fix.**
When a valid system code is found in the CSV, check if it's active/preview before adding to the structure that tracks subgraph inclusion. Easy!

Right now, it's *kind of* no-harm-no-foul, but [in the future](https://github.com/yakra/DataProcessing/issues/251), depending on how aggressive I want to get with conserving RAM, it may force our hand.

**errorcheck?**
Sure, these cases could be silently ignored. But so could unrecognized system codes, and there are errorchecks for those.

Having an errorcheck would let a contributor know that:
* the system code should be changed (e.g. typo, system got renamed, or something)
* or removed (both of these just the same as with unrecognized codes)
* or perhaps most importland... **WHY DO I MAKE THAT TYPO ALL THE TIME** *importantly*, the system should be promoted to preview in systems.csv.

### LINKED ISSUE #616: __builtin_ctz (count trailing zeros)
@jteresco, does this little program compile & run without issue on MacOS?
```c++
#include <iostream>

int main()
{	for (unsigned short n = 1; n; n <<= 1)
	  std::cout << n << " has " << __builtin_ctz(n) << " trailing 0(s).\n";
}
```

### LINKED ISSUE #621: System abbreviation error output
>  the errors given when two lines in systems.csv have the same system abbreviation are not very helpful at diagnosing the problem

https://github.com/TravelMapping/RailwayData/pull/140#issue-2257045450 by @neroute2

yakra:
* Closes #616.
* Closes #621.
* Closes #614.
* Closes yakra#269.
* Closes yakra#95.
* Closes yakra#265.
* Closes yakra#254.
* Closes yakra#251.

@jteresco, I'd recommend to be absolutely safe, downloading & doing a MacOS/XCode compile ([warnings](https://github.com/TravelMapping/DataProcessing/issues/585) notwithstanding) before merging this, due to using a few clang builtins.

### What's this about?
Taking care of a couple errorcheck items (#614, #621) along the way, the main focus here is implementing some long-planned speed improvements for graph generation.
A couple extra steps in the HighwayGraph ctor populate sets of vertex & edge pointers for each region & system in a subgraph.
When writing the subgraphs, we quickly populate vertex & edge sets via the magic of `|=` and (for fullcustom graphs) `&=`.
No more of [this clunky](https://github.com/TravelMapping/DataProcessing/blob/fd96ac64ba2eff23600e86df8dcc84e94cb0a916/siteupdate/cplusplus/classes/GraphGeneration/get_subgraph_data.cpp#L40-L64) iterating through vectors, chasing pointers around, checking membership with ad-hoc bitmasks, and populating more vectors.
Edges in particular benefit. There used to be [3 lists of incident edges](https://github.com/TravelMapping/DataProcessing/blob/fd96ac64ba2eff23600e86df8dcc84e94cb0a916/siteupdate/cplusplus/classes/GraphGeneration/HGVertex.h#L15-L17) for each vertex and [3 vectors of matching edges](https://github.com/TravelMapping/DataProcessing/blob/fd96ac64ba2eff23600e86df8dcc84e94cb0a916/siteupdate/cplusplus/classes/GraphGeneration/HighwayGraph.cpp#L281) for each subgraph. Simple, collapsed, traveled. Now, there's [1 vector](https://github.com/TravelMapping/DataProcessing/blob/32829211cd8f93bb26a76d8ac339e4c454168a55/siteupdate/cplusplus/classes/GraphGeneration/HGVertex.h#L16) and [1 TMBitset](https://github.com/TravelMapping/DataProcessing/blob/32829211cd8f93bb26a76d8ac339e4c454168a55/siteupdate/cplusplus/classes/GraphGeneration/HighwayGraph.cpp#L424).
No more redundantly checking an edge for membership 2, 4 or 6 times. It's *just there* in the TMBitset.
(The redundancy is limited to 2 checks, one for each vertex, back in the ctor when the region/system sets are populated.)
From here, we iterate 1 container, not 3, and write each edge's info to the simple, collapsed, or traveled files per its `format` bits.
HGVertex & HGedge objects are more lightweight, and they mey get smaller in the future too.

**What does it *not* do?**
This doesn't reach the goal of getting FreeBSD to scale gracefully to more threads ([next pull request](https://github.com/TravelMapping/DataProcessing/issues/585#issuecomment-2106346470)) but still yields a significant additional improvement.

### Nota Bene:
In d218c4d, [Region::ve_thread](https://github.com/TravelMapping/DataProcessing/commit/d218c4d7639d4370d9e8c44f59bacdf5495a91e4?diff=split&w=0#diff-725206f9efb720527dea3da0107c674f344f4dbe7649e00f5c86f4fb27f0d95bR76-R85) introduces some diffs to graphs, omitting 5 points in 4 graphs in 3 regions.
**TLDR:** When a location at a border contains only devel points in Region A and active/preview points only in Region B:
• **Before:** The vertex is included in A's region graph, and any 2+region graphs that include Region A but not B.
• **After:** The vertex is omitted from these graphs.
More details, including a list of the affected points & graphs, at https://github.com/yakra/DataProcessing/issues/265#issuecomment-2033378755.
I believe this is The Right Thing To Do.

### Misc. facts:
* I switched to the C++17 standard for a bit when some changes caused some weird behavior by the compiler & linking errors, and added in some C++17 syntactic sugar including structured bindings.
  Later, some last-minute optimizations removed the need to do this, so I reverted those commits, allowing us to continue to build siteupdate using older toolchains.
* 71 commits (69 if we don't count 2 that only reverted C++17 changes) got squashed down to 9.
* *Lambdas!* Lambdas are one of my new favorite things. [#617](https://github.com/TravelMapping/DataProcessing/pull/617) had the 1st one; [the 2nd](https://github.com/TravelMapping/DataProcessing/blob/fd96ac64ba2eff23600e86df8dcc84e94cb0a916/siteupdate/cplusplus/classes/TravelerList/TravelerList.cpp#L128-L139) was in [#619](https://github.com/TravelMapping/DataProcessing/pull/619).
  This pull request introduces 7 more:
  * [one to populate vertex & edge TMBitsets from a PlaceRadius](https://github.com/TravelMapping/DataProcessing/blob/32829211cd8f93bb26a76d8ac339e4c454168a55/siteupdate/cplusplus/classes/GraphGeneration/get_subgraph_data.cpp#L3-L11)
  * [one to detach temporary partially-collapsed edges from vertices during edge compression](https://github.com/TravelMapping/DataProcessing/blob/32829211cd8f93bb26a76d8ac339e4c454168a55/siteupdate/cplusplus/classes/GraphGeneration/HGEdge.cpp#L98-L104)
  * ...Then there's the new optional TMBitset logs. 5 new lambdas here, including 2 lambdas *within* a lambda because `niram` has only one edge in the entire system.
* TMBitset.cpp has tripled in size (that'd make a good [John Lennon](https://www.youtube.com/watch?v=EX7roXRmkOU) parody), 87 -> 261 lines.
* Getting all the HGEdge objects into contiguous storage ended up a lot simpler than [originally conceptualized](https://github.com/TravelMapping/DataProcessing/issues/518#issuecomment-1418233143). No need to make `new` objects on the heap & come up with clever ways to track & delete them if I can just emplace them in a temporary container like a `std::list` which goes out of scope at the end of the HighwayGraph ctor. I ultimately decided to do away with the temporary container and moves entirely, and just store all the edges, including the "dead" ones, in a TMArray. The cost of copying all the edges & then destroying the temporary container outweighed the benefit of keeping the HGEdge TMBitsets absolutely minimized and free of extraneous 0s. [TMBitset iteration is pretty efficient now](https://github.com/TravelMapping/DataProcessing/commit/a5c60b88efba28faa7522448ce812c29b70970bb#diff-f357887f8771cdd6d83cb932dd1b16acbc479e4fb8503a9f36bedf7d87fc4aefL55-R201), and the extra 0s don't hurt much.
* [I considered sorting](https://github.com/TravelMapping/DataProcessing/issues/518#issuecomment-2051026759) edges after construction in order to defragment their TMBitsets, but that took up more time (0.7 s on BiggaTomato) than it saved.
 Skipping the sorting optimizes for speed though not RAM use, and leaves open the possibility of finding clever ways to reorder edge construction in the future.

### DIFF
--- a/siteupdate/cplusplus/classes/Args/Args.cpp
+++ b/siteupdate/cplusplus/classes/Args/Args.cpp
@@ -7,6 +7,7 @@
 /* k */ bool Args::skipgraphs = 0;
 /* v */ bool Args::mtvertices = 0;
 /* C */ bool Args::mtcsvfiles = 0;
+/* E */ bool Args::edgecounts = 0;
 /* w */ std::string Args::datapath = "../../HighwayData";
 /* s */ std::string Args::systemsfile = "systems.csv";
 /* u */ std::string Args::userlistfilepath = "../../UserData/list_files";
@@ -33,6 +34,7 @@ bool Args::init(int argc, char *argv[])
 		else if ARG(0, "-k", "--skipgraphs")		 skipgraphs = 1;
 		else if ARG(0, "-v", "--mt-vertices")		 mtvertices = 1;
 		else if ARG(0, "-C", "--mt-csvs")		 mtcsvfiles = 1;
+		else if ARG(0, "-E", "--edge-counts")		 edgecounts = 1;
 		else if ARG(0, "-h", "--help")			{show_help(); return 1;}
 		else if ARG(1, "-w", "--datapath")		{datapath	  = argv[++n];}
 		else if ARG(1, "-s", "--systemsfile")		{systemsfile      = argv[++n];}
@@ -81,7 +83,7 @@ void Args::show_help()
 	std::cout  <<  indent << "        [-c CSVSTATFILEPATH] [-g GRAPHFILEPATH] [-k]\n";
 	std::cout  <<  indent << "        [-n NMPMERGEPATH] [-p SPLITREGIONPATH SPLITREGION]\n";
 	std::cout  <<  indent << "        [-U USERLIST [USERLIST ...]] [-t NUMTHREADS] [-e]\n";
-	std::cout  <<  indent << "        [-T TIMEPRECISION] [-v] [-C]\n";
+	std::cout  <<  indent << "        [-T TIMEPRECISION] [-v] [-C] [-E]\n";
 	std::cout  <<  indent << "        [-L COLOCATIONLIMIT] [-N NMPTHRESHOLD]\n";
 	std::cout  <<  "\n";
 	std::cout  <<  "Create SQL, stats, graphs, and log files from highway and user data for the\n";
@@ -124,6 +126,7 @@ void Args::show_help()
 	std::cout  <<  "		        timestamp readouts\n";
 	std::cout  <<  "  -v, --mt-vertices     Multi-threaded vertex construction\n";
 	std::cout  <<  "  -C, --mt-csvs         Multi-threaded stats csv files\n";
+	std::cout  <<  "  -E, --edge-counts     Report the quantity of each format graph edge\n";
 	std::cout  <<  "  -L, --colocationlimit COLOCATIONLIMIT\n";
 	std::cout  <<  "		        Threshold to report colocation counts\n";
 	std::cout  <<  "  -N, --nmp-threshold NMPTHRESHOLD\n";
--- a/siteupdate/cplusplus/classes/Args/Args.cpp
+++ b/siteupdate/cplusplus/classes/Args/Args.cpp
@@ -8,6 +8,7 @@
 /* v */ bool Args::mtvertices = 0;
 /* C */ bool Args::mtcsvfiles = 0;
 /* E */ bool Args::edgecounts = 0;
+/* b */ bool Args::bitsetlogs = 0;
 /* w */ std::string Args::datapath = "../../HighwayData";
 /* s */ std::string Args::systemsfile = "systems.csv";
 /* u */ std::string Args::userlistfilepath = "../../UserData/list_files";
@@ -35,6 +36,7 @@ bool Args::init(int argc, char *argv[])
 		else if ARG(0, "-v", "--mt-vertices")		 mtvertices = 1;
 		else if ARG(0, "-C", "--mt-csvs")		 mtcsvfiles = 1;
 		else if ARG(0, "-E", "--edge-counts")		 edgecounts = 1;
+		else if ARG(0, "-b", "--bitset-logs")		 bitsetlogs = 1;
 		else if ARG(0, "-h", "--help")			{show_help(); return 1;}
 		else if ARG(1, "-w", "--datapath")		{datapath	  = argv[++n];}
 		else if ARG(1, "-s", "--systemsfile")		{systemsfile      = argv[++n];}
@@ -83,7 +85,7 @@ void Args::show_help()
 	std::cout  <<  indent << "        [-c CSVSTATFILEPATH] [-g GRAPHFILEPATH] [-k]\n";
 	std::cout  <<  indent << "        [-n NMPMERGEPATH] [-p SPLITREGIONPATH SPLITREGION]\n";
 	std::cout  <<  indent << "        [-U USERLIST [USERLIST ...]] [-t NUMTHREADS] [-e]\n";
-	std::cout  <<  indent << "        [-T TIMEPRECISION] [-v] [-C] [-E]\n";
+	std::cout  <<  indent << "        [-T TIMEPRECISION] [-v] [-C] [-E] [-b]\n";
 	std::cout  <<  indent << "        [-L COLOCATIONLIMIT] [-N NMPTHRESHOLD]\n";
 	std::cout  <<  "\n";
 	std::cout  <<  "Create SQL, stats, graphs, and log files from highway and user data for the\n";
@@ -127,6 +129,8 @@ void Args::show_help()
 	std::cout  <<  "  -v, --mt-vertices     Multi-threaded vertex construction\n";
 	std::cout  <<  "  -C, --mt-csvs         Multi-threaded stats csv files\n";
 	std::cout  <<  "  -E, --edge-counts     Report the quantity of each format graph edge\n";
+	std::cout  <<  "  -b, --bitset-logs     Write TMBitset RAM use logs for region & system\n";
+	std::cout  <<  "		        vertices & edges\n";
 	std::cout  <<  "  -L, --colocationlimit COLOCATIONLIMIT\n";
 	std::cout  <<  "		        Threshold to report colocation counts\n";
 	std::cout  <<  "  -N, --nmp-threshold NMPTHRESHOLD\n";
--- a/siteupdate/cplusplus/classes/Args/Args.h
+++ b/siteupdate/cplusplus/classes/Args/Args.h
@@ -19,6 +19,7 @@ class Args
 	/* T */ static int timeprecision;
 	/* v */ static bool mtvertices;
 	/* C */ static bool mtcsvfiles;
+	/* E */ static bool edgecounts;
 	/* L */ static int colocationlimit;
 	/* N */ static double nmpthreshold; 
 		static const char* exec;
--- a/siteupdate/cplusplus/classes/Args/Args.h
+++ b/siteupdate/cplusplus/classes/Args/Args.h
@@ -20,6 +20,7 @@ class Args
 	/* v */ static bool mtvertices;
 	/* C */ static bool mtcsvfiles;
 	/* E */ static bool edgecounts;
+	/* b */ static bool bitsetlogs;
 	/* L */ static int colocationlimit;
 	/* N */ static double nmpthreshold; 
 		static const char* exec;
--- a/siteupdate/cplusplus/classes/GraphGeneration/GraphListEntry.cpp
+++ b/siteupdate/cplusplus/classes/GraphGeneration/GraphListEntry.cpp
@@ -6,11 +6,11 @@
 std::vector<GraphListEntry> GraphListEntry::entries;
 size_t GraphListEntry::num; // iterator for entries
 
-GraphListEntry::GraphListEntry(std::string r, std::string d, char f, char c, std::list<Region*> *rg, std::list<HighwaySystem*> *sys, PlaceRadius *pr):
+GraphListEntry::GraphListEntry(std::string r, std::string d, char f, char c, std::vector<Region*> *rg, std::vector<HighwaySystem*> *sys, PlaceRadius *pr):
 	regions(rg), systems(sys), placeradius(pr),
 	root(r), descr(d), form(f), cat(c) {}
 
-void GraphListEntry::add_group(std::string&& r, std::string&& d, char c, std::list<Region*> *rg, std::list<HighwaySystem*> *sys, PlaceRadius *pr)
+void GraphListEntry::add_group(std::string&& r, std::string&& d, char c, std::vector<Region*> *rg, std::vector<HighwaySystem*> *sys, PlaceRadius *pr)
 {	GraphListEntry::entries.emplace_back(r, d, 's', c, rg, sys, pr);
 	GraphListEntry::entries.emplace_back(r, d, 'c', c, rg, sys, pr);
 	GraphListEntry::entries.emplace_back(r, d, 't', c, rg, sys, pr);
--- a/siteupdate/cplusplus/classes/GraphGeneration/GraphListEntry.h
+++ b/siteupdate/cplusplus/classes/GraphGeneration/GraphListEntry.h
@@ -1,7 +1,6 @@
 class HighwaySystem;
 class PlaceRadius;
 class Region;
-#include <list>
 #include <string>
 #include <vector>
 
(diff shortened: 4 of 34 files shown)


yakra:
## Setting up for graphs of highway data

Starting at the start of graph generation....

Setup
Very small changes here; you can see we're near the margin of error, especially looking at lab2.
* `897c` was earlier in development and got rebased; it's equivalent to d218c4d.
  [We're no longer setting up the ad hoc bitfield for vertex membership.](https://github.com/TravelMapping/DataProcessing/commit/d218c4d7639d4370d9e8c44f59bacdf5495a91e4#diff-0a73788eb5d0f37cba3d1dd9c571f3760d4f99741935edfbdea647c3d0be9f9dL30-L33)
* By the 2nd-most-recent commit, [we're passing around & incrementing another variable, and placing larger structures into the hi_ & lo_priority_points vectors.](https://github.com/TravelMapping/DataProcessing/commit/86cfdb3fff2d76564905305410cc36fe143bcf38#diff-c045521fca87e318ca424d0aaa445af15b89af08da633432029f96c8ec293dd7)

yakra:
## Creating vertices and region & system vertex (and edge) sets

Vtx
Creating unique names and vertices, on its, own, gets a little faster.
[No calls to add_vertex, none of the vector appends & mutexes involved.](https://github.com/TravelMapping/DataProcessing/commit/d218c4d7639d4370d9e8c44f59bacdf5495a91e4#diff-851b8bd611a92bdb9ea361f0360bb49aaade90c0efd0c453816eaa94d2acfca2L24-L35)
*(Minor changes to [HighwayGraph::simplify](https://github.com/TravelMapping/DataProcessing/commit/86cfdb3fff2d76564905305410cc36fe143bcf38?diff=unified&w=0#diff-0a73788eb5d0f37cba3d1dd9c571f3760d4f99741935edfbdea647c3d0be9f9dL178) could in theory make it a little faster though in practice it's near margin of error.)*

However, this missing functionality has just been moved elsewhere.
Creating edge and vector sets now has separate multithreaded tasks for regions & systems.

rtmb
htmb
* The dotted lines represent a more apples-to-apples comparison, an earlier stage of development since squashed into d218c4d. Here, like no-build (`fd96`), only sets of vertices get populated.
* Edges take more time to do. We iterate through the incident edges of each vertex (several per), which means checking & adding each vertex twice, once for each vertex. Then we have to check each edge to ensure it's still in the right region/system, iterating through its concurrent segments if applicable. [...Now that I think of it, there may be a way to speed this up.](https://github.com/yakra/DataProcessing/issues/285)

VtxAll_6928
* Add in `6928`'s time for building only vertex sets and we can see that while there's more noise at more threads, with fewer threads we're still the tiniest bit ahead of the game.
  Except on bsdlab, where there's a more visible improvement.
* Nota bene: This doesn't include calling TMBitset::shrink_to_fit(), which may or may not push us a bit above no-build.
  I don't expect it has very much impact.

VtxAll
For comparison, here's the current combined time of all 3 tasks; creating vertices, region vertex & edge sets, and system vertex & edge sets.
You can see that with populating the edge sets, we're taking up more time now, but this is the upfront cost we pay for more speed when generating subgraphs.
Not the most relevant chart, though. We're measuring the sum of unrelated tasks. There were no per-region/system edge sets or vectors in fd96ac6.
Another way to see the impact of the edge sets is in the height difference between lines in per-region & per-system charts above.

Now let's rewind a bit for a fun little curiosity of vertex-only sets.
Recall the no-build creates HGVertex* vectors as opposed to the TMBitsets of `6928`.
Let's look at Phases 1 & 2 of [threaded set population](https://github.com/yakra/DataProcessing/issues/265).
* `2f63` = threaded population of vectors, vertices only.
* `6928` = threaded population of TMBitsets, vertices only.

rtmb_2f63
Regions:
* On the low end, vectors are more efficient.
  This makes sense, as adding a value to a TMBitset involves more arithmetic, bit-shifting & such.
* On the high end, TMBitsets are more efficient.
  This makes sense, as resizing a vector means memory copies & reallocations, which can add up.

htmb_2f63
Whereas for systems, vectors are faster across the board.
Hm. Far fewer systems (only those listed in a .csv as being in a subgraph) than regions get sets made, but that shouldn't matter; there will still be just as many threads doing copies & reallocations.
I can only speculate here, but maybe the time spent iterating through concurrent segments eases up the load on the memory bus just enough? \*Shrug.\*

yakra:
## Edge construction

Edge1
[The simple constructor is, well, simpler.](https://github.com/TravelMapping/DataProcessing/commit/d218c4d7639d4370d9e8c44f59bacdf5495a91e4#diff-f09ed1ba2938cce8cd6ab16188ce16a7821edb7a34612f75549942a7d0afdb0b)
* No more allocating & initializing the `written` bools.
* With each vertex now only having one container of incident edges
  rather than 3, we only push `this` edge to 2 containers rather than 6.
* Said container is now a std::vector rather than a std::list; that's worth something.

Edge2
The collapse ctor has even more improvements:
* Again, no setting up `written`.
* [No more conditional branching to find `edge1` and `edge2`; these are passed in from the HighwayGraph ctor now.](https://github.com/TravelMapping/DataProcessing/commit/d218c4d7639d4370d9e8c44f59bacdf5495a91e4#diff-f09ed1ba2938cce8cd6ab16188ce16a7821edb7a34612f75549942a7d0afdb0bL40-R31)
* With 1 incident edge vector per vertex now, we push `this` edge to [exactly 2](https://github.com/TravelMapping/DataProcessing/commit/d218c4d7639d4370d9e8c44f59bacdf5495a91e4#diff-f09ed1ba2938cce8cd6ab16188ce16a7821edb7a34612f75549942a7d0afdb0bR89-R90) containers rather than [potentially 4](https://github.com/TravelMapping/DataProcessing/commit/d218c4d7639d4370d9e8c44f59bacdf5495a91e4#diff-f09ed1ba2938cce8cd6ab16188ce16a7821edb7a34612f75549942a7d0afdb0bL109-L116), and lose some conditional branching.
* [HGEdge::detach is greatly simplified.](https://github.com/TravelMapping/DataProcessing/commit/d218c4d7639d4370d9e8c44f59bacdf5495a91e4#diff-f09ed1ba2938cce8cd6ab16188ce16a7821edb7a34612f75549942a7d0afdb0bL119-R103)
  * We erase the edge from [2 containers](https://github.com/TravelMapping/DataProcessing/commit/d218c4d7639d4370d9e8c44f59bacdf5495a91e4#diff-f09ed1ba2938cce8cd6ab16188ce16a7821edb7a34612f75549942a7d0afdb0bR102-R103) instead of [up to 6](https://github.com/TravelMapping/DataProcessing/commit/d218c4d7639d4370d9e8c44f59bacdf5495a91e4#diff-f09ed1ba2938cce8cd6ab16188ce16a7821edb7a34612f75549942a7d0afdb0bL128-L139).
  * `written` doesn't exist anymore. [No resources to free.](https://github.com/TravelMapping/DataProcessing/commit/d218c4d7639d4370d9e8c44f59bacdf5495a91e4#diff-f09ed1ba2938cce8cd6ab16188ce16a7821edb7a34612f75549942a7d0afdb0bL142-L145)
  * [It gets called less often.](https://github.com/TravelMapping/DataProcessing/commit/d218c4d7639d4370d9e8c44f59bacdf5495a91e4#diff-f09ed1ba2938cce8cd6ab16188ce16a7821edb7a34612f75549942a7d0afdb0bR87-R88) We only need to worry about erasing an edge from the one `HGVertex::incident_edges` when it "dies", when all its `format` bits are cleared.

Back at the HighwayGraph ctor, the [collapse condition](https://github.com/TravelMapping/DataProcessing/commit/d218c4d7639d4370d9e8c44f59bacdf5495a91e4?diff=split&w=0#diff-0a73788eb5d0f37cba3d1dd9c571f3760d4f99741935edfbdea647c3d0be9f9dL97-L100) (whether to do 1 dual-format collapse or 2 independent single-format collapses) is much simpler (yakra#269, yakra#95).
We get rid of some extraneous pointer-chasing too.

yakra:
## Writing graph files

And now, the good stuff!

Graphs
OK, so FreeBSD is a *little* bit faster much but not all of the time.
*Maybe* we're *starting* to get somewhere, but overall not much improvement. ...Yet.
The good news is, [I've found the magic bullet; the next pull request](https://github.com/TravelMapping/DataProcessing/issues/585#issuecomment-2106346470) will have bsdlab right down there hanging out with the Linux machines.

Graphs_Linux
Zoomed in to just the Linux machines.

yakra:
## Reading subgraph descriptions

Saving this for last *(well, out of the graph-related charts anyway)* because it's the least. :)

SubCsv
Yes, the Y-axis is in milliseconds. The bars on toma & lab3 suggest we're near margin of error.
* `e7b1` was earlier in development and got rebased; it's equivalent to e46619a.
  This uses [HighwaySystem::sysname_hash](https://github.com/TravelMapping/DataProcessing/issues/621#issuecomment-2080315837).
* As of 1350045, the subgraph region & system lists are replaced with vectors.

A few oddities & observations about the machines' relative performance:
* CentOS performs way worse than Ubuntu; lab3 & lab4 have 100% the same hardware.
  Lack of small string optimization, maybe?
* lab1 takes more time than lab2, despite its higher clock speed. I don't think its slower HD is a factor, as the benchmarks are done after a "dry run" of siteupdate that reads all the data it needs to into disk cache. Subsequent runs are much faster, in particular reading .wpt files. And I don't hear the HDs chugging away.
  More likely, its 2 RAM channels, as opposed to 3, are a factor.
* Contrast BiggaTomato (BT), which has an ancient HD and 2 channels of DDR2-800 (vice DDR3-1600) and performs way better because Ubuntu.
* I don't think bsdlab's SSD is a factor either.
  The only other hardware difference from lab3 & 4 is its RAM chips; they *may* have different timings.
  I wanna say the difference here is FreeBSD is just that much better at what it's doing.

yakra:
## Computing stats

One more last-minute pair of charts.

Stat
Computing stats also benefits from using `__builtin_ctzl` for TMBitset iteration.
...Ever so slightly.

Stat_log
More visible if I use a logarithmic scale? Maybe?

yakra:
*Ready to rock!*

jteresco:
It compiles, it runs datacheck.  7.2 seconds before, 5.5 seconds after.

I'll merge and use in the next site update.

jteresco:
This is on the production `siteupdate` instance (in my account) and on the production `datacheck instance` (in `/fast/tm`).

Thanks again!

jteresco:
Sure enough - both are about 5.5 second on second and third runs once things are cached.

michihdeu:
> It compiles, it runs datacheck. 7.2 seconds before, 5.5 seconds after.

I run data check  (/fast/tm) yesterday just after you wrote that. It took 12 seconds. I run it again now, and it was 14 seconds on first run and 12s on 2nd to 4th run. I wonder if it was already in place on /fast/tm when I run it yesterday?
btw, it's under 1s for railways

yakra:
^ I think the 5.5s figure is for Jim's M1 Mac rather than noreaster.

No surprise for railways; much smaller dataset! :grinning:

PR #186, opened 2019-02-05
check only edges incident to subgraph vertices

yakra:
Also contains the commits in #185.
Let's make sure that's good to go before merging this in.

Other than that, closes #179.

PR #623, opened 2024-05-19
#585 phase 1

### LINKED ISSUE #518: poor graph generation performance on FreeBSD

yakra:
Closes #518!

The performance-killing `sprintf` function has been removed from user log, DB file, and graph generation routines.
More to come in the future; I've yet to work on NMPs and stats CSVs.

@jteresco, I recommend a compile & test drive of this, if for no other reason than to see how blazingly fast graph generation is now. :grinning:

I am curious to see though, how fast  `--errorcheck` is with the userlog speedups.

One minor change: the `waypoints` table in the DB has done away with what I call "Python style floats", where a trailing `.0` is written after an integral value. `42.0` -> `42`.
This could in theory happen in other tables too; it's just in `waypoints` where we see it in practice. I don't expect it'll cause any problems.

---

We're nearing the end of the road for big **graph generation** improvements.
Percentagewise, this beats any of the [first](https://github.com/TravelMapping/DataProcessing/pull/180) [big](https://github.com/TravelMapping/DataProcessing/pull/182) [five](https://github.com/TravelMapping/DataProcessing/pull/185) [pull](https://github.com/TravelMapping/DataProcessing/pull/186) [requests](https://github.com/TravelMapping/DataProcessing/pull/193) from 2019. And it wasn't even something I had planned.
While I still have [a few ideas](https://github.com/yakra/DataProcessing/issues/270) left, there probably won't be anything else of this magnitude. Just smaller incremental changes, much lower impact.
I do expect to get lab2 to more consistently break the 1-second barrier, until the ever-increasing size of the HighwayData repo takes that possibility (using the most recent commit) away permanently.

### DIFF
--- a/siteupdate/cplusplus/classes/GraphGeneration/HGEdge.cpp
+++ b/siteupdate/cplusplus/classes/GraphGeneration/HGEdge.cpp
@@ -107,25 +107,21 @@ void HGEdge::detach()
 }
 
 // write line to tmg collapsed edge file
-void HGEdge::collapsed_tmg_line(std::ofstream& file, char* fstr, unsigned int threadnum, std::vector<HighwaySystem*> *systems)
+void HGEdge::collapsed_tmg_line(std::ofstream& file, unsigned int threadnum, std::vector<HighwaySystem*> *systems)
 {	file << vertex1->c_vertex_num[threadnum] << ' ' << vertex2->c_vertex_num[threadnum] << ' ';
 	segment->write_label(file, systems);
 	for (HGVertex *intermediate : intermediate_points)
-	{	sprintf(fstr, " %.15g %.15g", intermediate->lat, intermediate->lng);
-		file << fstr;
-	}
+		file << ' ' << intermediate->lat << ' ' << intermediate->lng;
 	file << '\n';
 }
 
 // write line to tmg traveled edge file
-void HGEdge::traveled_tmg_line(std::ofstream& file, char* fstr, unsigned int threadnum, std::vector<HighwaySystem*> *systems, bool trav, char* code)
+void HGEdge::traveled_tmg_line(std::ofstream& file, unsigned int threadnum, std::vector<HighwaySystem*> *systems, bool trav, char* code)
 {	file << vertex1->t_vertex_num[threadnum] << ' ' << vertex2->t_vertex_num[threadnum] << ' ';
 	segment->write_label(file, systems);
 	file << ' ' << (trav ? segment->clinchedby_code(code, threadnum) : "0");
 	for (HGVertex *intermediate : intermediate_points)
-	{	sprintf(fstr, " %.15g %.15g", intermediate->lat, intermediate->lng);
-		file << fstr;
-	}
+		file << ' ' << intermediate->lat << ' ' << intermediate->lng;
 	file << '\n';
 }
 
--- a/siteupdate/cplusplus/classes/GraphGeneration/HGEdge.h
+++ b/siteupdate/cplusplus/classes/GraphGeneration/HGEdge.h
@@ -31,8 +31,8 @@ class HGEdge
 	HGEdge(HGVertex *, unsigned char, HGEdge*, HGEdge*);
 
 	void detach();
-	void collapsed_tmg_line(std::ofstream&, char*, unsigned int, std::vector<HighwaySystem*>*);
-	void traveled_tmg_line (std::ofstream&, char*, unsigned int, std::vector<HighwaySystem*>*, bool, char*);
+	void collapsed_tmg_line(std::ofstream&, unsigned int, std::vector<HighwaySystem*>*);
+	void traveled_tmg_line (std::ofstream&, unsigned int, std::vector<HighwaySystem*>*, bool, char*);
 	std::string debug_tmg_line(std::vector<HighwaySystem*> *, unsigned int);
 	std::string str();
 	std::string intermediate_point_string();
--- a/siteupdate/cplusplus/classes/GraphGeneration/HighwayGraph.cpp
+++ b/siteupdate/cplusplus/classes/GraphGeneration/HighwayGraph.cpp
@@ -4,7 +4,6 @@
 #include "HGVertex.h"
 #include "PlaceRadius.h"
 #include "../Args/Args.h"
-#include "../Datacheck/Datacheck.h"
 #include "../ElapsedTime/ElapsedTime.h"
 #include "../HighwaySegment/HighwaySegment.h"
 #include "../HighwaySystem/HighwaySystem.h"
--- a/siteupdate/cplusplus/classes/GraphGeneration/HighwayGraph.cpp
+++ b/siteupdate/cplusplus/classes/GraphGeneration/HighwayGraph.cpp
@@ -345,9 +345,9 @@ void HighwayGraph::bitsetlogs(HGVertex* hp_end)
 //     for intermediate "shaping points" along the edge, ordered from endpoint 1 to endpoint 2.
 //
 void HighwayGraph::write_master_graphs_tmg()
-{	std::ofstream simplefile(Args::graphfilepath + "/tm-master-simple.tmg");
-	std::ofstream collapfile(Args::graphfilepath + "/tm-master.tmg");
-	std::ofstream travelfile(Args::graphfilepath + "/tm-master-traveled.tmg");
+{	std::ofstream simplefile(Args::graphfilepath + "/tm-master-simple.tmg");   simplefile.precision(15);
+	std::ofstream collapfile(Args::graphfilepath + "/tm-master.tmg");	   collapfile.precision(15);
+	std::ofstream travelfile(Args::graphfilepath + "/tm-master-traveled.tmg"); travelfile.precision(15);
 	simplefile << "TMG 1.0 simple\n";
 	collapfile << "TMG 1.0 collapsed\n";
 	travelfile << "TMG 2.0 traveled\n";
@@ -359,13 +359,14 @@ void HighwayGraph::write_master_graphs_tmg()
 	unsigned int sv = 0;
 	unsigned int cv = 0;
 	unsigned int tv = 0;
-	char fstr[57];
 	for (HGVertex& v : vertices)
-	{	sprintf(fstr, " %.15g %.15g", v.lat, v.lng);
+	{	auto write_vertex = [&](std::ofstream& tmg)
+		{	tmg << *(v.unique_name) << ' ' << v.lat << ' ' << v.lng << '\n';
+		};
 		switch (v.visibility) // fall-thru is a Good Thing!
-		{ case 2:  collapfile << *(v.unique_name) << fstr << '\n'; v.c_vertex_num[0] = cv++;
-		  case 1:  travelfile << *(v.unique_name) << fstr << '\n'; v.t_vertex_num[0] = tv++;
-		  default: simplefile << *(v.unique_name) << fstr << '\n'; v.s_vertex_num[0] = sv++;
+		{ case 2:  write_vertex(collapfile); v.c_vertex_num[0] = cv++;
+		  case 1:  write_vertex(travelfile); v.t_vertex_num[0] = tv++;
+		  default: write_vertex(simplefile); v.s_vertex_num[0] = sv++;
 		}
 	}
 
@@ -379,10 +380,10 @@ void HighwayGraph::write_master_graphs_tmg()
 	//TODO: multiple functions performing the same instructions for multiple files?
 	for (HGEdge *e = edges.begin(), *end = edges.end(); e != end; ++e)
 	{ if (e->format & HGEdge::collapsed)
-		e->collapsed_tmg_line(collapfile, fstr, 0, 0);
+		e->collapsed_tmg_line(collapfile, 0, 0);
 	  if (e->format & HGEdge::traveled)
 	  {	for (char*n=cbycode; n<cbycode+nibbles; ++n) *n = '0';
-		e->traveled_tmg_line(travelfile, fstr, 0, 0, TravelerList::allusers.size, cbycode);
+		e->traveled_tmg_line(travelfile, 0, 0, TravelerList::allusers.size, cbycode);
 	  }
 	  if (e->format & HGEdge::simple)
 	  {	simplefile << e->vertex1->s_vertex_num[0] << ' '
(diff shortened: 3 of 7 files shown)


yakra:
## userlogs

Here are links to charts of how [the](https://github.com/TravelMapping/DataProcessing/pull/607#issuecomment-1712928313) [last](https://github.com/TravelMapping/DataProcessing/pull/592#issuecomment-1627785857) [three](https://github.com/TravelMapping/DataProcessing/pull/526) performance improvements played out.
For FreeBSD in particular:
#526 helped out quite a bit at the high end, though a considerable bottleneck still remained.
#592 brought the low end way down, only to run right back into the same bottleneck.
#607 didn't have much effect.

User
This finally brings FreeBSD's time under control across the board, even outperforming both Linux flavors on the same hardware.

yakra:
## database

routes
connectedRoutes
waypoints
overallMileageByRegion
Note the Y axis in milliseconds.
At first the new times for toma & lab5 showed up as 0. I was able to tweak the shell script that extracts the times from the `siteupdate.log`s to get one more digit of precision, rather than re-run the bench mark with `--timeprecision 5`.

systemMileageByRegion
Milliseconds again.

clinchedOverallMileageByRegion
Interesting that toma has the worst performance, despite having the 3rd-highest clock speed.
Could it be that it has only 3 MB last-level cache vice 8+ for everything else?
Nope. There's one outlier of 0.0634s in its data; otherwise its average would be 0.0032s, better than all the other machines.

clinchedSystemMileageByRegion
clinchedConnectedRoutes
clinchedRoutes
db
Total time for the entire .sql file, including the tables not affected by these changes.

yakra:
## graph generation

This isn't new, but it should be noted that the .sql file is written in the background during the graph generation process.
This means that while *n* threads are dedicated to graph generation, the computer is running *n+1* threads overall, at least until the DB is finished or the subgraph threads start returning, whichever happens first.
Freeing up resources used by this background task gives us more memory bandwidth for graph generation, which of course itself is also sped up by `sprintf`'s removal.

Graphs_3282_110
Compare to the results from #622, the 1st image from [this post](https://github.com/TravelMapping/DataProcessing/pull/622#issuecomment-2111565402).
Now there's only one bsdlab line towering over all the rest at the top of the chart, the "before" line.

Graphs_3282_19
Zoom in and we can see that bsdlab's graph generation time is *finally* out of the stratosphere, down hanging out with all the Linux machines. It even outperforms lab3 & lab4 on the same hardware. **YES!**
A nice & relatively uncluttered chart too, with the slowest "after" lines well below the fastest "before" lines.

Graphs_3282_25
Same thing but with a 25s vertical scale, for easy side-by-side comparison with the zoomed-in chart from #622, the 2nd image from [this post](https://github.com/TravelMapping/DataProcessing/pull/622#issuecomment-2111565402).

I'd link to the images directly, but it looks like GitHub is adding some browser session-specific tokens to the URL.
If you want to manually open them in a new tab, gotta go to the post first. :(

yakra:
## alternate universe

Originally, this commit was done as a child of 3282921, but I decided to have a little fun with a merge commit.
2349b1f shows us...

### What if `sprintf` were removed first, instead of doing #622?

Graphs_2349-1_110
Again, only bsdlab's "before" line is way above the rest. The removal of `sprintf` was the magic bullet we needed to get the task to scale well.
Compare the 1st image from [this post](https://github.com/TravelMapping/DataProcessing/pull/622#issuecomment-2111565402).

Graphs_2349-1_25
Compare the 2nd image from [this post](https://github.com/TravelMapping/DataProcessing/pull/622#issuecomment-2111565402).
A much bigger impact than #622 on its own. FreeBSD is already performing better than Linux.
The "before" and "after" are still fairly separated, barely touching one another at 2 threads.
It'd appear that applying this change to #622 has a bigger impact than applying it to fd96ac6.

### And if #622 had been applied *after* this?

Graphs_2349-2_13
This gives a clearer picture of the benefits of using `TMBitset`s under more optimal circumstances.
Especially with FreeBSD, which actually has a meaningful delta now.

Graphs_2349-2_25
Same thing but with a 25s vertical scale again, if you wanna compare side-by-wide with whatever.

yakra:
Ready to rock!

jteresco:
Going to try it out now, thanks.

jteresco:
Under 30 seconds for a site update (minus DB loading) on my MBP.

yakra:
I'm curious to know how long it took before.
While MacOS has a BSD lineage, that doesn't necessarily mean it suffers the same memory bottlenecks as FreeBSD.
Another factor is that performance may not just depend on the OS, but also the under-the-hood implementation of `sprintf` in `stdio.h`, which could well be different on the Apple platform.

jteresco:
Hmmm..  Production site update went from 66 seconds to 112.

yakra:
Odd.
siteupdate.log lists 6223ec7b160bcb5158b68fea38f962f37d7bd088; that's the latest one.
It's not a disk cache thing; reading WPTs is finished @ 2.8s.

Could disk write access be the bottleneck?

Around the time of my last post, I noticed slow graph writes on bsdlab, which has an SSD.
Reproducing it now, I get 178.9s & 177.4s on 2 different runs.
Using `-g /dev/null`, I get 1.7s.

A few years back, I noticed widely varying results when benchmarking tasks that did a lot of writing to disk.
Same executable, same dataset, same number of threads.
Deleting lots of files would make these tasks run faster. Maybe the time depends on what disk sectors it's writing to? Maybe it's the OS coping with a fragmented filesystem, taking longer to find where it can store files?
Either way, I started writing all output files other that siteupdate.log itself to /dev/null when doing benchmarks, to get consistent & reproducible data (and to show what the software is capable of when disk access isn't a bottleneck :grin:).

* Are graph files written to SSD, or magnetic disk?
* What are your results with `-g /dev/null`?

michihdeu:
> The performance-killing `sprintf` function has been removed from user log, DB file, and graph generation routines. More to come in the future; I've yet to work on NMPs and stats CSVs.

I see that you've removed the function but don't get how you log now. Can you point me to the "new" function (call)?

michihdeu:
@yakra + @jteresco Something went wrong. Both production servers indicate **0 miles for all users** now

https://forum.travelmapping.net/index.php?topic=6238.msg34610#msg34610
https://travelmapping.net/stat.php
https://tmrail.teresco.org/stat.php

yakra:
\*puts fingers to keyboard to respond to previous post\*

...Crap.

yakra:
Anyway.
> I see that you've removed the function but don't get how you log now. Can you point me to the "new" function (call)?

As before, I'm using the overloaded [std::ostream::operator <<](https://cplusplus.com/reference/ostream/ostream/operator%3C%3C/).
(Overloaded operators are basically just functions with [syntactic sugar](https://en.wikipedia.org/wiki/Syntactic_sugar) on top.)
What's different now is how many times it's used and what gets inserted into the `ofstream`.

The changes in the code are too numerous to mention, but here's [one example from userlogs](https://github.com/TravelMapping/DataProcessing/pull/623/files#diff-8a323ef0bce4aba54cfd7eef56fdfeda18216c23c3f5135c13b00a89e5449386L115-R137):

**Before:** 
[A formatted string was prepared](https://github.com/TravelMapping/DataProcessing/pull/623/files#diff-8a323ef0bce4aba54cfd7eef56fdfeda18216c23c3f5135c13b00a89e5449386L115), "Traveled %i of %i (%.1f%%), Clinched %i of %i (%.1f%%) preview systems",
  where the bits starting with `%` were replaced by the 6 numbers on [the following 2 lines](https://github.com/TravelMapping/DataProcessing/pull/623/files#diff-8a323ef0bce4aba54cfd7eef56fdfeda18216c23c3f5135c13b00a89e5449386L116-L117).
  [This string was then inserted into the output file stream](https://github.com/TravelMapping/DataProcessing/pull/623/files#diff-8a323ef0bce4aba54cfd7eef56fdfeda18216c23c3f5135c13b00a89e5449386L118).

**After:**
[The components that made up that string are instead individually inserted into the ofstream](https://github.com/TravelMapping/DataProcessing/pull/623/files#diff-8a323ef0bce4aba54cfd7eef56fdfeda18216c23c3f5135c13b00a89e5449386R134-R137)...
Text << number << text << number << text << number << text << number << text << number << text << number << text.

yakra:
```diff
yakra@BiggaTomato:~/ytm/DataProcessing/siteupdate/cplusplus$ diff \
<(tail -n +12258954 _3282/TravelMapping.sql | head -n 3) \
<(tail -n +12258954 _spr1/TravelMapping.sql | head -n 3) \
# overallMileageByRegion table
1,3c1,3
< ('AB','9235.1470421475879','18472.872821377736')
< ,('ABW','45.461910136369539','45.461910136369539')
< ,('AFG','2493.3650900241528','2493.3650900241528')
---
> ('AB','0x1.20992d246ef83p+13','0x1.20a37dc4e3228p+14')
> ,('ABW','0x1.6bb1fdf10b2f4p+5','0x1.6bb1fdf10b2f4p+5')
> ,('AFG','0x1.37abaed1463aap+11','0x1.37abaed1463aap+11')
```

I use a custom program to diff the .SQL files, which apparently reads & interprets those hex values just fine, and converts them to a number internally.
MySQL on the other hand apparently doesn't like them so much.

Let's `git checkout` 52ef723 and use the old version on noreaster till I can get this sorted.

yakra:
Probably wanna nix `floatfield`. Me slwwp now. Sleep. Me slwwp... \*AHEM\* sleep now.

https://github.com/TravelMapping/DataProcessing/pull/623/files#diff-a9628ca98720ba21a19933f4b3b1059ed1d7c674a4b83988dce9bb160e448d7aR25

https://github.com/TravelMapping/DataProcessing/pull/623/files#diff-a9628ca98720ba21a19933f4b3b1059ed1d7c674a4b83988dce9bb160e448d7aR194

https://github.com/TravelMapping/DataProcessing/pull/623/files#diff-a9628ca98720ba21a19933f4b3b1059ed1d7c674a4b83988dce9bb160e448d7aR223

jteresco:
@yakra no urgency on this since it was easy enough to jump back to a previous point.

I'll try some things out with the file writes some time soon, don't hesitate to remind me if I don't get at it soon.

yakra:
[I know exactly what to do.](https://github.com/TravelMapping/DataProcessing/pull/623#issuecomment-2119771995)
Just going a little overboard on due diligence.
Also typing up a post-mortem in #624 on the unlikely series of very specific events got us here.

PR #628, opened 2024-06-04
{fmt}: #585 phase 1 take 2

yakra:
Closes #626.
This reverts & re-does #623.
The slowdown is gone, replaced with the speedup it really *should* have been.

Like before, `sprintf` has been removed from user log, DB file, and graph generation routines.
(And now, routedatastats.log & stats CSVs too.)
This time, it's been replaced with calls to the [{fmt}](https://fmt.dev) library instead of using iostreams for `double` formatting.

As with #622, I recommend downloading and doing a compile & test drive before merging:
* to make sure {fmt} is working on Mac
* OK, clang is working again on noreaster. See how fast graph generation is, for real this time!
* O HAY! How fast is `--errorcheck` now? :grin:

---

As noted before, we're nearing the end of the road for big graph generation improvements.
Percentagewise, this is in line with the [first](https://github.com/TravelMapping/DataProcessing/pull/180) [big](https://github.com/TravelMapping/DataProcessing/pull/182) [five](https://github.com/TravelMapping/DataProcessing/pull/185) [pull](https://github.com/TravelMapping/DataProcessing/pull/186) [requests](https://github.com/TravelMapping/DataProcessing/pull/193) from 2019.
It's about middle-of-the-pack, doesn't beat all of them hands-down. (#623 *did*, but only in the unrealistic scenario of writing to invalid files, e.g. /dev/null.)
While I still have [a few ideas](https://github.com/yakra/DataProcessing/issues/270) left, there probably won't be anything else of this magnitude. Just smaller incremental changes, much lower impact.
Although, now that we know formatting numbers is a hotspot, we can add a couple more items to that list:
* Experiment with `fmt::format_int` and {fmt}'s `std::ofstream` support.
  FreeBSD seems especially hampered by formatting numbers (at least, with `double`s); this may help.
* [Reduce redundancy (including formatting) when writing tmg edge lines.](https://github.com/yakra/DataProcessing/issues/286)

PR #632, opened 2024-06-14
bye-bye sprintf

### LINKED ISSUE #585: C++ siteupdate: Use of sprintf, deprecated on MacOS/XCode

yakra:
Closes #585.
Removes `sprintf` completely.
d4a8716 silences a few warnings from the new clang version on noreaster.

It's possible that XCode may still generate warnings for other functions deprecated for similar reasons, E.G. `styrcpy` used in the subgraph CSV routines.
Feel free to open another issue if this is the case.

"Python style floats", with a `.0` at the end of integral values, are being phased out.
Cleaner simpler code, I'm not worried about keeping things diffable to siteupdate.py now that it's deprecated, and these don't affect anything in practice anyway.
* `Waypoint::str()`
  No instances in nmpfps.log, so no problem.
* tm-master.nmp
  No diffs before <-> after.
* OUT_OF_BOUNDS & DUPLICATE_COORDS datachecks
  No instances in datacheckfps.log, so no problem.

Meanwhile, the one affecting the `waypoints` sql table was phased out in #623 & phased back in in #628.
\*shrug\* Eh, it keeps the .sql files diffable?

disk-logs
Hm. I expected more of a difference when writing to [/dev/null](https://github.com/TravelMapping/DataProcessing/assets/13720877/9ba29993-dc19-4cf4-8d2d-67b67a93a475).
There's lots of copying, comparing & sorting strings, and checking set membership.

disk-merge1
What -- lab2 & lab3 are *worse* now? This looks like the [disk access bottleneck variability](https://github.com/TravelMapping/DataProcessing/pull/623#issuecomment-2119599132) that made me start writing to /dev/null. So, I deleted mountains of logs stats & graphs left over from previous runs and tried again.
disk-merge2
"After" mostly improves on "before" this time around, but we get there by way of stuff performing worse. Go figure.

This may be one of those rare cases when the /dev/null chart is more informative.
null-merge_12 0
FreeBSD improves tons. Disk access wasn't the bottleneck; it was `sprintf`.
Let's zoom in.
null-merge_3 5
Everything scales up well, even if bsdlab lags behind the others.
Lab3 is the fastest on its hardware now, and lab2 is the fastest overall.

### DIFF
--- a/siteupdate/cplusplus/classes/ElapsedTime/ElapsedTime.cpp
+++ b/siteupdate/cplusplus/classes/ElapsedTime/ElapsedTime.cpp
@@ -1,20 +1,15 @@
+#define FMT_HEADER_ONLY
 #include "ElapsedTime.h"
+#include <fmt/format.h>
 
 ElapsedTime::ElapsedTime(int precision)
 {	start_time = std::chrono::steady_clock::now();
-	format = "[%.1f] ";
-	format[3] = '0' + precision;
-	str = new char[15+precision];
-	      // deleted by ~ElapsedTime
+	format = "[{:.1f}] ";
+	format[4] = '0' + precision;
 }
 
 std::string ElapsedTime::et()
 {	using namespace std::chrono;
 	duration<double> elapsed = duration_cast<duration<double>>(steady_clock::now() - start_time);
-	sprintf(str, format.data(), elapsed.count());
-	return str;
-}
-
-ElapsedTime::~ElapsedTime()
-{	delete[] str;
+	return fmt::format(format.data(), elapsed.count());
 }
--- a/siteupdate/cplusplus/classes/ElapsedTime/ElapsedTime.h
+++ b/siteupdate/cplusplus/classes/ElapsedTime/ElapsedTime.h
@@ -5,10 +5,8 @@ class ElapsedTime
 {	//To get a nicely-formatted elapsed time string for printing
 	std::chrono::steady_clock::time_point start_time;
 	std::string format;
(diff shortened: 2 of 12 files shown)


yakra:
Ready to rock.

jteresco:
This will be used for tonight's site update.  The common datacheck version in `/fast/tm` is also now updated.

PR #634, opened 2024-06-26
OUT_OF_BOUNDS bugfix

yakra:
an actual OUT_OF_BOUNDS instance would have
```
terminate called after throwing an instance of 'fmt::v10::format_error'
  what():  argument not found
```


yakra:
Ready to rock!

yakra:
Merged.
Suggest using this for production site updates immediately.
If not, and any OUT_OF_BOUNDS cases creep into the data, the old siteupdate will crash as noted in the OP.

jteresco:
Sorry, didn't get it in for tonight, but it will be in use tomorrow.

PR #638, opened 2024-08-21
fix 6-field error precedence for devel systems

yakra:
https://travelmapping.net/updates/2024-08-18@21:57:32/logs/users/ssoworld.log
> ...
Ignoring line matching highway in system in development: IA LinHwy US30_Whe IA LinHwy CRY44
Ignoring line matching highway in system in development: IA LinHwy US30_GM IA LinHwy US30/61
IA LINHWY and IL LINHWY not in same connected route in line: IA LinHwy US30_DW IL LinHwy US30_Ste
Ignoring line matching highway in system in development: IL LinHwy IL40 IL LinHwy US52_S
Ignoring line matching highway in system in development: IL LinHwy IL251_S IL LinHwy IL251_N
...

This changes the middle error to another `Ignoring line matching route(s) in system in development` error.
That's the more important error here; as-is, it could kinda give the impression that we'd be able to mark travels on these routes if not for the ConnectedRoute error.

---

Along the way, changes "highway" to "route" in userlog errors, as we're expanding beyond highways now.

### DIFF
--- a/siteupdate/cplusplus/classes/TravelerList/mark_chopped_route_segments.cpp
+++ b/siteupdate/cplusplus/classes/TravelerList/mark_chopped_route_segments.cpp
@@ -13,7 +13,7 @@ if (rit == Route::pri_list_hash.end())
 		  {	*c = '?';
 			invalid_char = 1;
 		  }
-		log << "Unknown region/highway combo in line: " << trim_line;
+		log << "Unknown region/route combo in line: " << trim_line;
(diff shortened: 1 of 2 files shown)


yakra:
Having second thoughts about this.
* Makes processing of valid 6-field entries less efficient.
* Once the system goes preview, the entry will need to be fixed anyway, because different ConRoutes.
* Depending on how hidden system implementation shakes out, this part pf the code may change. May as well start from no-build.

PR #639, opened 2024-08-28
cross-region AltRouteNames

yakra:
https://forum.travelmapping.net/index.php?topic=6412


jteresco:
Any reason not to start using this right away?

Issue #636, opened 2024-07-04
Traveled-format graphs should exclude unranked travelers


jteresco:
The list of travelers and the bit fields on graph edges indicating travels should exclude unranked travelers introduced in #71.

PR #640, opened 2024-09-16
listfileinfo.csv robustness improvements

yakra:
Earlier vesions of 178644a & c00e9c6 were ancestors of my hidden system work in progress.
I hit pause on that after discovering the crashes that c38c5bd fixes, in order to work on robustness & get this stuff sorted.
The hidden system commits have been rebased on top of these ones.

A walk through the commits...

### 178644a variable name cleanup
De minimis readability improvements.
Being anal for the sake of being anal, really. :)

### c38c5bd failsafe traveler defaults
A quality-of-life improvement for me as a developer.
When working with older UserData commits with no listfileinfo.csv, these *"default defaults"* allow me to use an empty file as a placeholder.
Without them, writing the listEntries DB table would attempt to read past the end of the `TravelerList::defaults` vector, and crash.
Now, with an empty file I can write a valid DB. A *Missing* file = an `ErrorList` error & thus skipping the DB. More on this below.

### 4b71d2a strcspn-based readfields lambda
* More robust in handling CSVs with just one column (ie, just the listnames).
  Before, the one column would get re-parsed as a phantom 2nd column, as the 1st datum. Fixed.
* Using the C-string `strscpn` function simplifies & cleans up the `for` loops.
  It returns the length of a field, whether it ends at the delimiter or null terminator, letting us remove `next_delim` entirely.
  Plus it brings the `c` and `spn` variables in line with other uses program-wide -- `c` + `spn` = `strcspn`! :)
* In addition, a lambda cuts down on boilerplate & keeps all that stuff in one place.
* Removed a few extraneous scope resolution operators.

### c00e9c6 listfileinfo.csv not found
Originally, this just added the full path to the error text when the file was missing.
Now, proceeds through the rest of the function, with the variables/lambdas/loops all doing nothing until the "default defaults" are assigned.
This may be more useful in the future if we do more with this data than just dump it into the DB. (E.G., #636?)
For now, we have all the info necessary *to* write a valid DB; we just don't because there's an ErrorList error. More on *this* below.

### 385f8e8 std::move keys/values to listinfo
Replaced a bunch of memory copies with a couple moves, though in practice it won't save enough time to notice. :)
An errorcheck for multiple entries for the same user was super easy to do at the same time, so in it goes, even if it's not the most useful thing.

### 0493474 trim trailing whitespace & skip blank lines
`terescoj;Jim's lifetime travels;1`
`terescoj;Jim's lifetime travels;1 `
See the difference? The former would put TRUE in the DB as expected; the latter FALSE.
Trying to trim *leading* whitespace messed with my existing code & lambdas, so I'm not bothering with it. It's way easier to spot in a text editor than trailing anyway.
It's common for blank lines to make their way into a file; skipping them prevents `Number of fields does not match number of fieldnames` errors.

# *The question is...*
Should a missing listfileinfo.csv still be an errorcheck error?
* **No?** [If fullcustom.csv can't be opened, siteupdate skips right over that bit and writes a valid DB, no harm no foul.](https://github.com/TravelMapping/DataProcessing/blob/d4a8716/siteupdate/cplusplus/tasks/subgraphs/fullcustom.cpp#L3) Same deal here.
  I'd like to be able to freely checkout & work with older repos without having to temporarily modify the code or create a dummy file.
* **Yes?** All I can really think of here is, it lets us know if a bad commit accidentally deletes an existing file or changes permissions or whatever. But how likely is that?

@jteresco, thoughts?
I'd kinda like to make a missing file not an error.
If that's OK, then I'll do some rebasing & get all this done in one less commit.

### DIFF
--- a/siteupdate/cplusplus/classes/Args/Args.cpp
+++ b/siteupdate/cplusplus/classes/Args/Args.cpp
@@ -12,7 +12,7 @@
 /* w */ std::string Args::datapath = "../../HighwayData";
 /* s */ std::string Args::systemsfile = "systems.csv";
 /* u */ std::string Args::userlistfilepath = "../../UserData/list_files";
-/* x */ std::string Args::userlistextension = ".list";
+/* x */ std::string Args::userlistext = ".list";
 /* d */ std::string Args::databasename = "TravelMapping";
 /* l */ std::string Args::logfilepath = ".";
 /* c */ std::string Args::csvstatfilepath = ".";
@@ -42,7 +42,7 @@ bool Args::init(int argc, char *argv[])
 		else if ARG(1, "-w", "--datapath")		{datapath	  = argv[++n];}
 		else if ARG(1, "-s", "--systemsfile")		{systemsfile      = argv[++n];}
 		else if ARG(1, "-u", "--userlistfilepath")	{userlistfilepath = argv[++n];}
-		else if ARG(1, "-x", "--userlistextension")	{userlistextension = argv[++n];}
+		else if ARG(1, "-x", "--userlistext")		{userlistext	  = argv[++n];}
 		else if ARG(1, "-d", "--databasename")		{databasename     = argv[++n];}
 		else if ARG(1, "-l", "--logfilepath")		{logfilepath      = argv[++n];}
 		else if ARG(1, "-c", "--csvstatfilepath")	{csvstatfilepath  = argv[++n];}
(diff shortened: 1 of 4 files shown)


yakra:
## TLDR:

This makes it so that a valid DB can still be produced if listfileinfo.csv cannot be opened.
Shall I go ahead and remove the ErrorList check for this?

jteresco:
I am ok with the site update completing properly when there is no listfileinfo.csv.

yakra:
Commits have been rebased & reordered:
* **178644a variable name cleanup** Literal same commit as in the OP.
* **strcspn-based readfields lambda** Same diff as in the OP.
* **failsafe traveler defaults** Same diff as in the OP, except a missing file writes a DB & completes properly.
* **std::move keys/values to listinfo** Same diff as in the OP.
* **trim trailing whitespace & skip blank lines**  Same diff as in the OP.

jteresco (commented):
I didn't notice anything in the changes that seemed concerning to me.

PR #643, opened 2024-09-28
TMBitset::shrink_to_fit empty set bugfix

### LINKED ISSUE #642: railway graph generation crashes
```sh
make -j 4 && ./siteupdateST -x .rlist -t 4 -d .//TravelMapping -l .//logs -c .//stats -g .//graphs -u /home/yakra/tm/UserData/rlist_files -w /home/yakra/tm/RailwayData | tee .//logs/siteupdate.log; echo
```
```
RailwayData @ 87b0c6831fab496d2c8fd296969c895fe44136f9
UserData @ dd2fb13c2842000e0484939d24a2e04a51ae9e4e
DataProcessing @ e772c7175a61c83fbbaf7b41ef50b23b2bfaa122
```
siteupdate crashes when getting to the area graphs. Either it's the first one, siena2, or area graphs are bad news altogether.
@jteresco, did you have any problems with this too?

**Edit:** Crash reproduced on both bsdlab & noreaster using the current HEAD,
same RD & UD commits as the [current site update](https://tmrail.teresco.org/updates/2024-09-26@22:37:18/rlogs/siteupdate.log), which ran OK.
That update used 37c5ee3c2fda58c33a7ce17994193c8db74ce114 (only 1 commit since then), which *also crashes* when I attempt it.
What's the location of the binary on noreaster used for production updates?
Does it have a `diff` from /home/yakra/ytm/DataProcessing/siteupdate/cplusplus/siteupdate_37c5?

This may or may not be related:
`ERROR: segment name mismatch in HGEdge collapse constructor: edge1 named BLp,BLp,NSLn edge2 named NSLn`
Related: #91 #178 #360 #589 #603
* [x] Do newer commits make the error and/or crash disappear? **No.**
* ~If so, what data changes that?~
* [x] Other than that, need to understand the intersection between the segment name mismatch & the crashes, cause & effect, before proceeding

yakra:
closes #642

### DIFF
--- a/siteupdate/cplusplus/templates/TMBitset.cpp
+++ b/siteupdate/cplusplus/templates/TMBitset.cpp
@@ -77,8 +77,10 @@ template <class item, class unit> class TMBitset
 
 	void shrink_to_fit()
 	{	if (!len) return;
+		size_t lo_index = *ibegin();
+		if (lo_index == len) return clear();
+		lo_index &= ~(ubits-1);
 		unit* const old_data = data;
-		size_t lo_index = *ibegin() & ~(ubits-1);
 
 		start += lo_index;
 		len = end_index() - lo_index;


yakra:
https://github.com/TravelMapping/DataProcessing/issues/642#issuecomment-2380266500
> Looks like @jteresco did know about problems with graphs, if `git status` is any indication:
> ```
> [yakra@noreaster /fast/home/terescoj/travelmapping/RailwayData/graphs]$ git status
> On branch master
> Your branch is up to date with 'origin/master'.
> 
> Changes to be committed:
>   (use "git restore --staged <file>..." to unstage)
> 	deleted:    areagraphs.csv
> 
> Changes not staged for commit:
>   (use "git add/rm <file>..." to update what will be committed)
>   (use "git restore <file>..." to discard changes in working directory)
> 	deleted:    multiregion.csv
> 	deleted:    systemgraphs.csv
> 
> Untracked files:
>   (use "git add <file>..." to include in what will be committed)
> 	../hwy_data
> ```

@jteresco, some questions & comments:

### CSVs changed on noreaster, listed above:
* **systemgraphs.csv**
  I'd like to add this back in, with just usanycs for now.
* **areagraphs.csv**
  This caused the #642, and can now be added back in, with the caveat that many graphs (e.g. siena2) will have no edges, or nothing at all.
  Since the RailwayData version was created, HighwayData has had entries added for worcester, ccu & whitman.
* **multiregion.csv**
  Was there an issue with this one that led to you deleting it?
  I could see (many of?) these also being relevant to rail.

### The other CSVs:
* **multisystem.csv**
  Nice, I see there's been a lot of effort put into putting together interesting datasets. :)
* **fullcustom.csv**
  No entries here. Fair enough, probably nothing out there relevant to rail transport.
  Might have been cleaner to not include this file at all. No header, no nothin'.
  I designed it a bit differently from the others, for siteupdate to move right along without complaining if the file isn't opened, as the fullcustom graphs were "silently" implemented for a while before fullcustom.csv was added.
  With a header-only file with no entries, a graph category gets added to the DB, `graph_types.push_back({"fullcustom", "Full Custom Graphs", "These graphs can be restricted by any combination of one more more regions and one or more highway systems, and optionally within the given distance radius of a given place."});`, even if there are no actual graphs *in* this category. Probably no-harm-no-foul, but it just doesn't seem quite right. :)
  With no file at all OTOH, this empty category is not added.

**And on that note...**
* As for the other CSVs, whether they're header-only or don't exist at all, they'll also add a graph category entry. There's a built-in assumption that these files are populated.
  I guess the most consistent & robust thing to do would be, for all 5 CSVs, to track whether any GraphListEntries are added, and add the graph category entry based on that.

Issue #626, opened 2024-05-21
Graph writing speeds


jteresco:
Graph writing appears to have slowed significantly on FreeBSD with the changes from #623 

With the current C++ siteupdate (5/21/2024), 8 threads, on FreeBSD, times for graph generation (computation and file writes):

Writing to the `/fast` SSD: ~86 sec
Writing to the `/home` magnetic drive: ~98 sec
Writing to `/ramdisk` RAM disk: ~96 sec

(only one run each for now)

But it doesn't seem to matter much where I write.  It also looks like the 8 threads are all pegging their cores during the graph generation process.

yakra:
Oh boy. All this after I claimed it closed #518.
Mind if I log into noreaster & see what I can see?

yakra:
https://travelmapping.net/updates/2024-05-21@15:12:31/logs/siteupdate.log
looks like to one done right before the OP here.
> DataProcessing @ 2cc40e097919a510f1c800ff7c14842e2f68c8d1

checks out; indeed the most recent revision.

WRT the approximate times, I use the following timestamps when measuring graph speed:
```
[14.8] Writing master TM graph files.
[107.1] Clearing HighwayGraph contents from memory.
```
so I get 92.3 s.

yakra:
Compiled executables for 32829211cd8f93bb26a76d8ac339e4c454168a55 (AKA 52ef723da83485741366bcf6afdefe448dc49e69) & 2cc40e097919a510f1c800ff7c14842e2f68c8d1,
`siteupdate_3282` and `siteupdate_2cc4`.

Running
`/home/yakra/ytm/DataProcessing/siteupdate/cplusplus/check10.sh -h NEer -n -tt 8 -tp 4 -p 2 -r Graphs _3282 _2cc4`
and the same thing *(minus the `-h NEer` bit)* concurrently on bsdlab.
When done I'll do the same adding the `--null` option, which writes to /dev/null.

yakra:
Reproducing these results so far.
I suppose it's *possible* that... the `std::ofstream` class can detect when it has a file in `/dev/null` "open", and skip the formatting of inserted doubles *(or \*anything\* really)* as a time-saver?

And maybe my [writing to /dev/null](https://github.com/TravelMapping/DataProcessing/pull/623#issuecomment-2119599132) isn't as representative of the real world as I'd been thinking all this time?

[...]

PR #681, opened 2026-02-01
graph generation speedups (three different ones)

### LINKED ISSUE #255: vertex numbers
Carving out this side topic in order to close #217. Some background/notes there.

A few different **v7** & **v8** flavors are on BiggaTomato in branches:
**v7gs** `5849d90`
**v8** `d8da55e`
Refamiliarize.

---

Another possible **v7** variant...
```c++
// simple, collapsed & traveled vertex numbers, per thread
int& HighwayGraph::s_vertex_num(HGVertex const* v, const int threadnum)
{	return vnums[(threadnum*vertices.size()+v-vertices.data())*3];
}
int& HighwayGraph::c_vertex_num(HGVertex const* v, const int threadnum)
{	return vnums[(threadnum*vertices.size()+v-vertices.data())*3+1];
}
int& HighwayGraph::t_vertex_num(HGVertex const* v, const int threadnum)
{	return vnums[(threadnum*vertices.size()+v-vertices.data())*3+2];
}
```
That's a hell of a lot of arithmetic to get our array indices.
* [ ] Convert `vnums` to a `static thread_local` HGVertex member.
  This gets rid of `threadnum*vertices.size()+`, `size()` itself being calculated via `begin()-end()`.
* [ ] Save an index to the first pointer: `unsigned int vnum = HGVertex::vnums+(v-vertices.data())*3;`
* [ ] Get our refs via `vnum[0]`, `vnum[1]` & `vnum[2]`.

### LINKED ISSUE #589: nix HGEdge::segment_name

yakra:
Closes yakra#255
Closes #589, even though it'd be fun to try out 32-byte edges at some point

These graph generation speedups finally open up the FreeBSD bottleneck, bringing its performance in line with Linux.
Noreaster is gonna love this.

What's going on under the hood is discussed in separate posts for each of the 3 commits below.
Here are the charts for how performance is affected by all 3 commits in this pull request combined.

4-f2f9
4-f2f9-log

The second image is the same thing on a logarithmic scale.
This can help more easily differentiate the before & after lines and the different machines, with everything less squished together in the lower left corner.
It also helps more easily visualize where the sweet spot is for each machine, before efficiency starts decreasing.

Benchmarks are performed using a RAM disk now.
For those who remember the #626 fiasco (which led to adopting {fmt} in the 1st place), this avoids both the inconsistently slow times that can result from writing to disk, and the falsely fast times recorded when writing to /dev/null.

Never thought I'd see this, but we're just shy of breaking the 1 second barrier.
For graph generation proper (not counting the now-separate "Formatting vertex coordinate strings" task), lab6 averages 1.0707 s @ 7 threads. Individual passes have taken as little as 1.0191 s (6 threads in that case).
I have a few more tweaks in the pipeline long-term that can get us there -- at least, as long as I use old enough HighwayData & UserData revisions. At some point the data will increase to the point that sub-1s graph generation will be permanently out of reach.
At some point returns will diminish and there will be no more efficiency to wring out of the process.

### DIFF
--- a/siteupdate/cplusplus/classes/GraphGeneration/HGEdge.cpp
+++ b/siteupdate/cplusplus/classes/GraphGeneration/HGEdge.cpp
@@ -100,7 +100,9 @@ void HGEdge::detach()
 // write line to tmg collapsed edge file
 void HGEdge::collapsed_tmg_line(std::ofstream& file, char* fstr, unsigned int threadnum, std::vector<HighwaySystem*> *systems)
 {	file << vertex1->c_vertex_num[threadnum] << ' ' << vertex2->c_vertex_num[threadnum] << ' ';
-	segment->write_label(file, systems);
+	if (systems)
+		segment->write_label(file, systems);
+	else	file << segment_name;
 	for (HGVertex *intermediate : intermediate_points)
 	{	*fmt::format_to(fstr, " {:.15} {:.15}", intermediate->lat, intermediate->lng) = 0;
 		file << fstr;
@@ -111,7 +113,9 @@ void HGEdge::collapsed_tmg_line(std::ofstream& file, char* fstr, unsigned int th
 // write line to tmg traveled edge file
 void HGEdge::traveled_tmg_line(std::ofstream& file, char* fstr, unsigned int threadnum, std::vector<HighwaySystem*> *systems, bool trav, char* code)
 {	file << vertex1->t_vertex_num[threadnum] << ' ' << vertex2->t_vertex_num[threadnum] << ' ';
-	segment->write_label(file, systems);
+	if (systems)
+		segment->write_label(file, systems);
+	else	file << segment_name;
 	file << ' ' << (trav ? segment->clinchedby_code(code, threadnum) : "0");
 	for (HGVertex *intermediate : intermediate_points)
 	{	*fmt::format_to(fstr, " {:.15} {:.15}", intermediate->lat, intermediate->lng) = 0;
--- a/siteupdate/cplusplus/classes/GraphGeneration/HGEdge.cpp
+++ b/siteupdate/cplusplus/classes/GraphGeneration/HGEdge.cpp
@@ -98,25 +98,21 @@ void HGEdge::detach()
 }
 
 // write line to tmg collapsed edge file
-void HGEdge::collapsed_tmg_line(std::ofstream& file, char* fstr, unsigned int threadnum, std::vector<HighwaySystem*> *systems)
+void HGEdge::collapsed_tmg_line(std::ofstream& file, unsigned int threadnum, std::vector<HighwaySystem*> *systems)
 {	file << vertex1->c_vertex_num[threadnum] << ' ' << vertex2->c_vertex_num[threadnum] << ' ';
 	segment->write_label(file, systems);
 	for (HGVertex *intermediate : intermediate_points)
-	{	*fmt::format_to(fstr, " {:.15} {:.15}", intermediate->lat, intermediate->lng) = 0;
-		file << fstr;
-	}
+		file << intermediate->coordstr;
 	file << '\n';
 }
 
 // write line to tmg traveled edge file
-void HGEdge::traveled_tmg_line(std::ofstream& file, char* fstr, unsigned int threadnum, std::vector<HighwaySystem*> *systems, bool trav, char* code)
+void HGEdge::traveled_tmg_line(std::ofstream& file, unsigned int threadnum, std::vector<HighwaySystem*> *systems, bool trav, char* code)
 {	file << vertex1->t_vertex_num[threadnum] << ' ' << vertex2->t_vertex_num[threadnum] << ' ';
 	segment->write_label(file, systems);
 	file << ' ' << (trav ? segment->clinchedby_code(code, threadnum) : "0");
 	for (HGVertex *intermediate : intermediate_points)
-	{	*fmt::format_to(fstr, " {:.15} {:.15}", intermediate->lat, intermediate->lng) = 0;
-		file << fstr;
-	}
+		file << intermediate->coordstr;
 	file << '\n';
 }
 
--- a/siteupdate/cplusplus/classes/GraphGeneration/HGEdge.cpp
+++ b/siteupdate/cplusplus/classes/GraphGeneration/HGEdge.cpp
@@ -97,29 +97,6 @@ void HGEdge::detach()
 	detach(vertex2->incident_edges);
 }
 
-// write line to tmg collapsed edge file
-void HGEdge::collapsed_tmg_line(std::ofstream& file, char* fstr, unsigned int threadnum, std::vector<HighwaySystem*> *systems)
-{	file << vertex1->c_vertex_num[threadnum] << ' ' << vertex2->c_vertex_num[threadnum] << ' ';
-	segment->write_label(file, systems);
-	for (HGVertex *intermediate : intermediate_points)
-	{	*fmt::format_to(fstr, " {:.15} {:.15}", intermediate->lat, intermediate->lng) = 0;
-		file << fstr;
-	}
-	file << '\n';
-}
-
-// write line to tmg traveled edge file
-void HGEdge::traveled_tmg_line(std::ofstream& file, char* fstr, unsigned int threadnum, std::vector<HighwaySystem*> *systems, bool trav, char* code)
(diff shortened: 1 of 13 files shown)


yakra:
Let's actually hold off on this one a wee bit.
There's a minor code organization thing I'd like to look into for e76ad220622aebbd75ae84cd701c829b21ab0bc0.

yakra:
Hm, noreaster was actually doing pretty well *before* this update!

```
[yakra@noreaster /home/www/tm/updates]$ egrep '^\[[0-9]+\.[0-9]\]' 2026-01-31@21:29:46/logs/siteupdate.log | egrep 'master TM|Clearing HighwayGraph'; echo
[8.9] Writing master TM graph files.
[13.0] Clearing HighwayGraph contents from memory.

[yakra@noreaster /home/www/tm/updates]$ egrep '^\[[0-9]+\.[0-9]\]' 2026-02-01@21:19:37/logs/siteupdate.log | egrep 'master TM|Clearing HighwayGraph'; echo
[7.8] Writing master TM graph files.
[11.0] Clearing HighwayGraph contents from memory.

[yakra@noreaster /home/www/tm/updates]$ echo 13.0 - 8.9 | bc
4.1
[yakra@noreaster /home/www/tm/updates]$ echo 11.0 - 7.8 | bc
3.2
[yakra@noreaster /home/www/tm/updates]$ clang++ --version
FreeBSD clang version 19.1.7 (https://github.com/llvm/llvm-project.git llvmorg-19.1.7-0-gcd708029e0b2)
Target: x86_64-unknown-freebsd14.3
```

Could be that the newer OS & clang version have taken care of what was going on under the hood that held FreeBSD back relative to Linux.
```
[yakra@bsdlab ~]$ clang++ --version
FreeBSD clang version 14.0.5 (https://github.com/llvm/llvm-project.git llvmorg-14.0.5-0-gc12386ae247c)
Target: x86_64-unknown-freebsd13.2
```
Sure, noreaster does have a higher clock speed, more cache, and what, 4-channel DDR4? (Compared to dual Xeon E5-2420s.) But I doubt that alone could explain its performance, considering how poorly BSDlab does vs Linux on identical hardware...
How long has it been running FreeBSD 14.3?

yakra:
```
[yakra@noreaster /home/www/tm/updates/monthly_snapshots]$ for d in 2024-06-01@22:28:25 2024-07-01@21:52:28 2024-08-01@22:45:39 2024-09-01@22:32:17 2024-10-01@21:33:39 2024-11-01@22:15:07 2024-12-01@22:12:20 2025-01-01@21:48:33 2025-02-02@11:02:57 2025-03-01@22:35:01 2025-04-01@21:23:31 2025-05-01@22:05:09 2025-06-01@22:03:15; do echo -en "$d\t\t"; (echo -n 0-; egrep '^\[[0-9]+\.[0-9]\]' $d/logs/siteupdate.log | egrep 'master TM|Clearing HighwayGraph' | sed -e 's~[^0-9.]~~g' -e 's~\.$~+~' | tr -d '\n'; echo 0) | bc; done
2024-06-01@22:28:25		89.7
2024-07-01@21:52:28		11.8
2024-08-01@22:45:39		11.9
2024-09-01@22:32:17		13.0
2024-10-01@21:33:39		12.6
2024-11-01@22:15:07		12.3
2024-12-01@22:12:20		11.6
2025-01-01@21:48:33		12.2
2025-02-02@11:02:57		4.3
2025-03-01@22:35:01		3.9
2025-04-01@21:23:31		3.9
2025-05-01@22:05:09		5.1
2025-06-01@22:03:15		4.1
```
June-July 2024 coincides with #628.
What happened last February?

PR #682, opened 2026-02-01
error check for duplicate graph names

yakra:
```
[17.7] ABORTING due to 1 errors:
1: Duplicate graph name Tourist in multisystem graphs
```
If a graph name is duplicated,
* single-threaded graph generation overwrites the 1st graph with the 2nd. The 1st has a DB entry, but no file.
* multithreaded graph generation attempts to write both graphs to the same file simultaneously, resulting in a corrupted file.

This prevents that from happening.
Ping @si404


si404:
Thanks Eric, and welcome back.

Fixed the error locally. Pull coming in shortly

